PR #48, opened 2019-02-27
Fix array handling

philjdf:


csharp.test/TestLogicalTypeRoundtrip.cs
```diff
@@ -149,6 +149,44 @@ public static void TestBigArrayRoundtrip()
             }
         }
 
+        [Test]
+        public static void TestAwkwardArrayRoundtrip()
```
GPSnoopy: Awkward => EdgeCases

csharp/LogicalColumnWriter.cs
```diff
@@ -106,136 +108,143 @@ public override void WriteBatch(ReadOnlySpan<TElement> values)
...
                 WriteBatchSimple(values, converter as LogicalWrite<TElement, TPhysical>.Converter);
             }
         }
 
-        private void WriteArrayIntermediateLevel(
-            ReadOnlySpan<TElement> values,
-            short repetitionLevel, short maxRepetitionLevel, short leafFirstRepLevel, 
-            short[] nullDefinitionLevels, short leafDefinitionLevel, 
-            LogicalWrite<TLogical, TPhysical>.Converter converter)
+        private static List<Node> GetSchemaNode(Node node)
```
GPSnoopy: Duplicated with reader, belongs in base class.

Also, as discussed, done and checked at ctor time so it does not need to be called at each read.

csharp/LogicalColumnStream.cs
```diff
@@ -50,40 +38,18 @@ private static (short nestingDepth, short[] nullDefinitionLevels) GetArraySchema
...
-            {
-                throw new Exception("Schema does not match type we are trying to read into.");
-            }
+        protected readonly Array Buffer;
+        protected readonly short[] DefLevels;
+        protected readonly short[] RepLevels;
 
-            return (checked((short) nestingDepth), nullDefinitionLevels);
-        }
+        protected Schema.Node[] ArraySchemaNodes;
```
GPSnoopy: readonly ?

Issue #50, opened 2019-03-08
Expose ResizableBuffer initial size ctor argument


SpencerBullent:

GPSnoopy:
Also check if Buffer itself can be constructed with a fixed size.

GPSnoopy:
Fixed by PR #51.

PR #51, opened 2019-03-29
Expose initial size for resizable buffer

philjdf:
https://github.com/G-Research/ParquetSharp/issues/50

### DIFF
--- a/cpp/ResizableBuffer.cpp
+++ b/cpp/ResizableBuffer.cpp
@@ -6,7 +6,7 @@
 
 extern "C"
 {
-	PARQUETSHARP_EXPORT ExceptionInfo* ResizableBuffer_Create(std::shared_ptr<arrow::ResizableBuffer>** buffer, int64_t initialSize)
+	PARQUETSHARP_EXPORT ExceptionInfo* ResizableBuffer_Create(int64_t initialSize, std::shared_ptr<arrow::ResizableBuffer>** buffer)
 	{
 		TRYCATCH(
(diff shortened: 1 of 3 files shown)


cpp/ResizableBuffer.cpp
```diff
@@ -6,11 +6,11 @@
 
 extern "C"
 {
-	PARQUETSHARP_EXPORT ExceptionInfo* ResizableBuffer_Create(std::shared_ptr<arrow::ResizableBuffer>** buffer)
+	PARQUETSHARP_EXPORT ExceptionInfo* ResizableBuffer_Create(int64_t initialSize, std::shared_ptr<arrow::ResizableBuffer>** buffer)
```
GPSnoopy: const?

csharp/ExceptionInfo.cs
```diff
@@ -10,6 +10,7 @@ namespace ParquetSharp
     internal sealed class ExceptionInfo
     {
         public delegate IntPtr GetAction<TValue>(out TValue value);
+        public delegate IntPtr GetAction<TValue, TArg0>(TArg0 arg0, out TValue value);
```
GPSnoopy: Swap TArg0 and TValue to be consistent with GetFunction<in TArg0, TValue> below (also notice the in keyword)?

cpp/ResizableBuffer.cpp
philjdf: const.

csharp/ExceptionInfo.cs
philjdf: Swapped.

PR #58, opened 2019-07-16
Version 2.0.0 upgraded to Arrow 0.14

GPSnoopy:
Breaks compatibility due to changed API around LogicalType.
Added nanosecond support via TimeSpanNanos and DateTimeNanos.
Added support for LogicalType.Int8 and Int16 (sbyte, byte, short, ushort).

TODO: raise issue with Arrow for Decimal statistics being set, but meaningless values.


philjdf:
I have a branch of on my vcpkg fork, Arrow014andThriftStatic, which combines the Thrift update with the Arrow 0.14 upgrade.

PR #76, opened 2019-11-05
Close streams when Managed* classes are deleted

Frassle:
Add "leaveOpen" parameter to opt out of this (same parameter name as
StreamWriter/Reader).

### DIFF
--- a/cpp/ManagedOutputStream.cpp
+++ b/cpp/ManagedOutputStream.cpp
@@ -4,6 +4,7 @@
 
 #include <arrow/status.h>
 #include <arrow/io/interfaces.h>
+#include <arrow/util/logging.h>
 
 using arrow::Status;
 using arrow::StatusCode;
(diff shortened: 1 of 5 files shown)


cpp/ManagedOutputStream.cpp
```diff
@@ -34,6 +35,10 @@ class ManagedOutputStream final : public arrow::io::OutputStream
 
 	~ManagedOutputStream()
 	{
+		arrow::Status st = this->Close();
+		if (!st.ok()) {
+			ARROW_LOG(FATAL) << "Error ignored when destroying ManagedOutputStream: " << st;
```
herebebeasties: Should this be FATAL if it's ignored and we carry on regardless? ERROR perhaps?

Frassle: Yes it should be ERROR.

GPSnoopy:
I'll have some time to work on ParquetSharp end of next week. We've got a PR update to vcpkg Arrow that has been accepted, the latest releases of Arrow fixes quite a few Parquet bugs.

GPSnoopy:
Being merged into master as part of PR #77. Nuget package already built: https://www.nuget.org/packages/ParquetSharp/2.0.2-beta2

PR #79, opened 2019-11-12
Compression enum values changed in 0.15.x :-\

GPSnoopy:


GPSnoopy:
They've consolidated Parquet and Arrow enums to be the same. Given that Arrow is 0.x, I guess it comes with little guarantees of backward compatibility.

It's lucky I even noticed, I was adding a new enum entry for Bz2.

PR #80, opened 2019-11-13
Fix a use after free bug

Frassle:
If there we're no managed reference left to a ManagedOutputStream or ManagedRandomAccessFile the class (and all it's delegates) could be collected while there was still a native reference to it from the C++ ParquetFileReader or ParquetFileWriter.

This PR adds a GCHandle to both managed class that keeps them alive until Free is called, which is then called by the C++ destructor.

### DIFF
--- a/cpp/ManagedOutputStream.cpp
+++ b/cpp/ManagedOutputStream.cpp
@@ -14,6 +14,7 @@ typedef StatusCode(*TellFunc)(int64_t*, const char**);
 typedef StatusCode(*FlushFunc)(const char**);
 typedef StatusCode(*CloseFunc)(const char**);
 typedef bool (*ClosedFunc)();
+typedef void (*FreeFunc)();
 
 class ManagedOutputStream final : public arrow::io::OutputStream
 {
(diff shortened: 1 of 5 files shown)


GPSnoopy:
Note that in order to trigger this bug, one has not to dispose properly of the ManagedOutputStream/ManagedRandomAccessFile instances. Otherwise the managed code still has a reference to them, waiting to call Dispose() on them.

The unit tests were arguably slightly incorrect, as the file stream was long disposed of before the parquet reader/writer were create. I'm not sure how this even worked. I have updated this as part of 
d731a85.

GPSnoopy:
PR #81 will merge this to master.

Frassle:
The file streams weren't destroyed as the managed class still held a reference to them

GPSnoopy:
Something is still bothering me about the original testing code. The ManagedOutputStream and it's ParquetHandle are clearly disposed of before the the writer is returned. The original code should not work as this not a valid thing (a lot of the C# classes in the .NET Framework tend to throw an InvalidOperationException when you use a disposed object). I think we're just being lucky here.

```
private static ParquetFileWriter MakeWriter()
{
    using (var input = new ManagedOutputStream(File.OpenWrite("file.parquet")))
    {
        return new ParquetFileWriter(input, new Column[] {new Column<int>("ids")});
    }
}
```

Frassle:
It's because the OutputStream is put into a pointer to shared pointer and dispose deletes that pointer, but the native code has a copy of the shared pointer as well so the actual OutputStream isn't deleted till the native code also deletes its shared pointer. Its not the clearest thing to follow.

PR #81, opened 2019-11-15
Fix early GC collect of managed IO wrapper classes

GPSnoopy:


GPSnoopy:
I believe this works 'cause of the smart pointer on the C++ side. Need to double check.

GPSnoopy:
I think this can cause memory to leak the other way around. Unless the C# instance is properly disposed of, the Free() method never gets called as the circular reference keeps the GC from collection and finalizing the IO classes.

Putting a break point or an exception in the Free method never seem to get hit. Further confirming the above suspicions.

I think in this case the cure is worse than the disease. Doing a proper dispose pattern in C# is easier than fixing the lifetime problem in a robust way.

PR #84, opened 2019-11-26
Fixed compression argument being ignored by ParquetRowWriter.

GPSnoopy:
KeyValueMetaData was also being ignored.


GPSnoopy:
Also bumped release version to 2.0.2-beta4

PR #100, opened 2020-02-05
Avoid duplicate builds on pull requests

jgiannuzzi:
Builds will still be triggered by pull requests.
For pushes, they will only be triggered if it's a push to the master branch.


GPSnoopy:
For the moment, the CI seems to be run all the time, duplicated for both push and PR. I'm willing to give this a shot and revert/change it if it's not testing enough.

PR #123, opened 2020-04-22
Introduce new NuGet CI workflow

GPSnoopy:
- Builds Linux and Windows native library separately.
- Builds and runs .NET Core unit tests separately as well, just to confirm all is well.
- Pass the native libraries as artifacts.
- Final .NET Core build to create nuget package and upload it as an artifact.

~~It's a bit unfortunate that the badges cannot point to a specific jobs. So we can either remove the Linux and Windows CIs and have a single badge, or keep them to build them concurrently with the Nuget CI. Not decided yet until I get more experience with this new GitHub Actions workflow.~~

Decided to get rid of the old Windows and Linux CI. Feedback per platform/job in GitHub is pretty good.


.github/workflows/nuget.yml
```diff
@@ -0,0 +1,83 @@
...
+      with:
+        name: linux-native-library
+        path: bin
+    - name: Download native ParquetSharp library (Windows)
+      uses: actions/download-artifact@v1
+      with:
+        name: windows-native-library
+        path: bin
+    - name: Build & Run .NET unit tests
+      run: dotnet test csharp.test --configuration=Release
```
jgiannuzzi, line 78: We could maybe use `dotnet pack` here instead of testing again on Windows?

GPSnoopy: Yes, indeed. Given that the native library is built by a separate job, I wanted to double check that everything is okay by re-running the .NET unit tests.

Compared to the vcpkg dependencies, this stage should be fairly quick. So I think it's worth the added safety. It's only redundant if we didn't fuck it up 😅

PR #124, opened 2020-04-23
Use GitHub Actions cache to avoid recompiling vcpkg dependencies every time.

GPSnoopy:
Consolidate vcpkg tag/commit-id in vcpkg_version.txt.
Use GitHub Actions cache to avoid recompiling vcpkg dependencies every time.

Note 100% reliable (see https://github.com/actions/cache/issues/191) but does provide a substantial boost when it works (e.g. Linux job completed in 2 minutes instead of 30 minutes). Worst case: it's as slow as before.

Things to be improved: make the cache key reflect the git commit-id, and consolidate the vcpkg git URL (@jgiannuzzi has kindly volunteered).


jgiannuzzi:
2 remarks:
- I think we are currently missing a step to write to the cache after building vcpkg
- We might want to enforce (at least through documentation) that vcpkg_version.txt should not contain a branch name, otherwise the caching will be done wrong

jgiannuzzi:
As a safeguard, we might want to use the vcpkg commit ID as the cache key, instead of a hash of the version. This would ensure that even using a branch name in vcpkg_version.txt would end up with proper caching.

GPSnoopy:
> I think we are currently missing a step to write to the cache after building vcpkg

My understanding is that the writing to the cache is implicit at the end of the job. It's "magic" ( I hate it).

> we might want to use the vcpkg commit ID as the cache key

How would you do that? My main problem is that `key: vcpkg-linux-${{ hashFiles('vcpkg_version.txt') }}` seems to have access to only a limited set of builtin functions. I haven't found a way to call any arbitrary bash command.

GPSnoopy:
Happy to have an extra check to ensure that vcpkg_version.txt contains a valid tag or commit-id only. Any suggestions?

jgiannuzzi:
> My understanding is that the writing to the cache is implicit at the end of the job. It's "magic" ( I hate it).

I had a closer look and I understand it that way too. Same hate for magic here 😡

jgiannuzzi:
> > we might want to use the vcpkg commit ID as the cache key
> 
> How would you do that? My main problem is that `key: vcpkg-linux-${{ hashFiles('vcpkg_version.txt') }}` seems to have access to only a limited set of builtin functions. I haven't found a way to call any arbitrary bash command.

I am thinking of having a step that "converts" the vcpkg version to a commit ID, and then we can use the output of that step in the cache key. I can figure out how to do that conversion if you'd like.

PR #131, opened 2020-04-25
Port to macOS and unify native build CI jobs

jgiannuzzi:
This is a port to macOS, with CI support and all unit tests working. The README states that this port is unsupported, as it is there mostly for convenience of development, not for actual production work.


.github/workflows/ci.yml
```diff
@@ -99,13 +99,58 @@ jobs:
         name: windows-native-library
         path: bin/ParquetSharpNative.dll
         
-  # Download both native shared libraries,
-  # Rebuild .NET projects (will automatically pick up both libraries and create a nuget package).
+  # Build and test everything on macOS.
+  # Upload the native shared library as an artifact.
+  native-osx:
```
GPSnoopy, line 104: native-osx and native-linux look very similar. I wonder if they can be matrixed (with fail-safe = false) to reduce complexity here. I think we agree that the Windows one is too different.

CMakeLists.txt
```diff
@@ -47,6 +46,12 @@ if (UNIX)
 
 endif ()
 
+if (CMAKE_HOST_SYSTEM_NAME STREQUAL "Darwin")
+  add_compile_options("-std=c++11")
+elseif (CMAKE_HOST_SYSTEM_NAME STREQUAL "Darwin")
```
GPSnoopy, line 51: Looks very suspicious that both conditions are equal. Also if OSX is considered to be UNIX by CMake, I would rather keep it in the UNIX block above for readability.

README.md
```diff
@@ -126,6 +126,12 @@ For building Arrow (including Parquet) and its dependencies, we recommend using
 > ./build_linux.sh
 > dotnet build csharp.test --configuration=Release
 ```
+**macOS (unsupported)**
```
GPSnoopy, line 129: Not sure if there is any value to claim it as unsupported at this point, given that it's CIed as everything else. Maybe Experimental would be a better warning?

My main worry is more around future upgrades of Arrow, and whether the MacOOS branch is going to delay a release... We'll see

GPSnoopy: Also, MacOS rather than macOS for consistency?

CMakeLists.txt
jgiannuzzi: Bad copy/paste! The goal was to move those linker flags for Linux only. I will fix it soon.

.github/workflows/ci.yml
jgiannuzzi: Good point, I will try to implement it that way.

README.md
jgiannuzzi: Good point too, I went with experimental at first, and then changed it to unsupported for that exact reason. It’s not experimental, as the unit tests work. And it is unsupported, as we don’t support it. Maybe alpha or beta?

For the name, the official one is macOS (don’t shoot the messenger!)

.github/workflows/ci.yml
GPSnoopy: I meant fail-fast BTW, not fail-safe. As blogged about here: https://www.edwardthomson.com/blog/github_actions_6_fail_fast_matrix_workflows.html

README.md
GPSnoopy: Maybe nothing at all then? It'll all depend whether macOS gets in the way of future releases for GR or not really.

macOS instead of MacOS, you're right I've just checked Wikipedia. o_O

jgiannuzzi:
Example of a CI run without cache: https://github.com/jgiannuzzi/ParquetSharp/actions/runs/88544344

Example of a CI run with cache:
https://github.com/jgiannuzzi/ParquetSharp/actions/runs/88564347

jgiannuzzi:
Review comments have been addressed, and the native build jobs have been unified (thanks to the magic of bash on Windows). See individual commits for more details.

CMakeLists.txt
```diff
@@ -37,13 +38,13 @@ endif ()
...
-	set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -s -Wl,--no-undefined")
 
 	add_compile_options("-Wall")
-	add_compile_options("-fvisibility=hidden")
 	add_compile_options("-fvisibility-inlines-hidden")
-	# add_link_options("-s")
-	# add_link_options("-Wl,--no-undefined")
+
+  if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
+    add_link_options("LINKER:--no-undefined")
```
GPSnoopy, line 46: This is not the same as before (missing `-s`, and potentially `-Wl` - I guess that the latter is replaced by `LINKER:`?)

`add_link_options` is not support by the CMake version that comes with Ubuntu 18.04. I guess that the GitHub image has a more recent version, but it will not work on a local computer with a standard Ubuntu install. Hence why it was commented out before.

cpp/CMakeLists.txt
```diff
@@ -105,3 +105,11 @@ target_link_libraries(ParquetSharpNative
 add_definitions(-DARROW_STATIC)
 add_definitions(-DARROW_NO_DEPRECATED_API)
 add_definitions(-DPARQUET_STATIC)
+
+if (UNIX AND CMAKE_BUILD_TYPE STREQUAL Release)
+  if (CMAKE_HOST_SYSTEM_NAME STREQUAL "Darwin")
+    set(STRIP_OPTIONS -x)
```
GPSnoopy, line 111: Not great that the `-s` lives in the mainfile while this is her. Probably best to move the `-s` here.

CMakeLists.txt
jgiannuzzi: `-s` is not portable and is used to strip the symbols. We are now using the `strip` command in a post build step instead. I confirm that the output of `-s` and `strip` is identical on a vanilla Ubuntu 18.04.

Good point about `add_link_options`, I reverted it in the latest commit.

cpp/CMakeLists.txt
jgiannuzzi: See explanation above.

PR #132, opened 2020-04-28
Test built NuGet package on each platorm during CI

jgiannuzzi:


.github/workflows/ci.yml
```diff
@@ -107,12 +105,51 @@ jobs:
       with:
         name: macOS-native-library
         path: bin
-    - name: Build & Run .NET unit tests
-      run: dotnet test csharp.test --configuration=Release
     - name: Build NuGet package
-      run: dotnet pack csharp --configuration=Release
+      run: |
+        dotnet build csharp --configuration=Release
```
GPSnoopy, line 110: According to the doc, `dotnet pack` implies `dotnet build`.

.github/workflows/ci.yml
```diff
@@ -82,9 +82,7 @@ jobs:
         name: ${{ runner.os }}-native-library
```
GPSnoopy, line 82: For some reason, GitHub does not allow to comment the code above your PR. Nonetheless, I think now that the `dotnet test` at line 78 is redundant (as well as the step name). We should probably replace it with `dotnet build`.

.github/workflows/ci.yml
jgiannuzzi: I started with only `dotnet pack` and got an error about missing `ParquetSharp.dll`, which is why I added `dotnet build`.

After your comment, I just checked with .NET Core 3.1.201 on macOS, and it seems that it's the opposite: `dotnet build` builds the assembly and creates the NuGet package, whilst `dotnet pack` does not build the assembly.

Do you agree to replace the above with only `dotnet build`?

.github/workflows/ci.yml
jgiannuzzi: Good point. Maybe we go one step further and only build the native library? I think there is no reason to build the .NET assembly anymore, and it would also respect the job name (`build-native`).

.github/workflows/ci.yml
GPSnoopy: You're right. Dotnet seems to have trouble with the native component when doing `dotnet pack`, despite the documentation clearly stating that it should do a build.

Fine, let's go for `dotnet build`, which seems to work for me locally as well.

.github/workflows/ci.yml
GPSnoopy: Was wondering the same thing.

It has the advantages of checking whether the whole repository compiles out of the box. Because we do quite a few manipulations on `csharp.test` in the `test-nuget` job, it's quite possible for the whole repository to be broken for local builds and we would never notice (e.g. missing dependencies on `csharp`, build order, etc).

I.e. this tests the developer workflow. While the jobs below test the user (nuget) worflow.

jgiannuzzi: Agreed, let's keep it as-is then.

.github/workflows/ci.yml
jgiannuzzi: Alright, will change that in the next commit.

PR #134, opened 2020-05-14
Add ParquetSharp.targets to the buildTransitive folder too

saul:
This ensures ParquetSharpNative.dll is copied to all transitive consumers of this package.


GPSnoopy:
After talking to @saul, it appears that the current transitivity works fine in .NET Core but not when mixing .NET Standard and .NET Framework. Documentation appears to confirm that buildTransisitive is what we want. I don't claim to be an expert.

PR #138, opened 2020-06-01
Fix vcpkg_windows.bat for case when GITHUB_ACTIONS is unset

asjflondon:


vcpkg_windows.bat
```diff
@@ -7,7 +7,7 @@ mkdir build
 cd build || goto :error
 git clone %vcpkg_url% -b %vcpkg_ref% vcpkg.windows || goto :error
 cd vcpkg.windows || goto :error
-if %GITHUB_ACTIONS%==true echo set(VCPKG_BUILD_TYPE release) >> triplets\x64-windows-static.cmake || goto :error
+if "%GITHUB_ACTIONS%"==true echo set(VCPKG_BUILD_TYPE release) >> triplets\x64-windows-static.cmake || goto :error
```
jgiannuzzi, line 10: Hi @asjflondon, thanks for your PR!

Can you elaborate on why this fix is needed?
I just tested on Windows 10 and `if %UNDEFINED% == value` does not fail.

However, `if "%DEFINED%" == value` is never true, as it's missing `"` around `value`, so this PR would make the CI build the vcpkg dependencies in debug and release, instead of release-only.

asjflondon: Thanks, that's odd. The batch file didn't work for me giving an error
`(echo was unexpected at this time.` (Win10, x64 Native Tools Command Prompt for VS 2019). I'll close this and next time I need to build see what happens.

PR #140, opened 2020-06-01
 Fix vcpkg_windows.bat for case when GITHUB_ACTIONS is unset (Take 2)

asjflondon:
Revised version of https://github.com/G-Research/ParquetSharp/pull/138 after comments
I think without the quotes it's not valid bat file syntax
`if %NOT_EXISTING_VAR%==true echo Hello`
works for me on the commandline but not in a batch file
http://steve-jansen.github.io/guides/windows-batch-scripting/part-5-if-then-conditionals.html


GPSnoopy:
I can confirm that `vcpkg_windows.bat` wasn't working before this PR, and that the problem is now fixed.

GPSnoopy:
Thanks for the contribution @asjflondon  :-)

PR #149, opened 2020-08-27
Add support for commit SHA as ref in vcpkg_version.txt

jgiannuzzi:
This PR adds the ability to directly target a commit SHA in `vcpkg_version.txt`.

It can be useful when support for a new version of Arrow gets added to vcpkg, and we neither want to wait for a release of vcpkg, nor do we want to use `master`, as it could change at any time.

### DIFF
--- a/vcpkg_unix.sh
+++ b/vcpkg_unix.sh
@@ -32,8 +32,10 @@ EOF
 
 mkdir -p build
 cd build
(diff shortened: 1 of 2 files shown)


jgiannuzzi:
Validation CI run can be found here: https://github.com/jgiannuzzi/ParquetSharp/actions/runs/226870861

GPSnoopy (approved):

PR #156, opened 2020-10-08
Fix vcpkg cache in CI workflow

jgiannuzzi:
Trying to resolve the vcpkg version of a commit SHA was giving an empty string, which was then causing the caching key to be wrong:

|      | Example |
| ---- | --- |
| Bad  | image |
| Good | image |

This PR fixes that by first checking whether we already have a commit SHA in `vcpkg_version.txt`, and only resolving if it's a tag/branch name instead.


GPSnoopy (approved):
Had to do some bash documentation reading. Guess things could go wrong if we use a 40 lettersvcpkg tag at one point? ;-)

PR #157, opened 2020-11-17
Sort fields and properties by MetadataToken.

jhickson:
I found an issue that when writing parquet files using a writer like:

```c#
var writer =
    ParquetFie.CreateRowWriter<(long, int, double, int, long, short)>(
        "path/to/my.parquet",
        new [] {"C1", "C2", "C3", "C4", "C5", "C6"}
    );
```

the resulting parquet file would have column C1 typed as `int` and C2 as `long` - i.e. the first two column types were swapped.

Upon debugging it transpired that this was when `Type.GetFields()` was being used to build the write delegate. `Type.GetFields()` was not returning fields in declaration order so the mapping of column name to type failed. The .NET docs for this method (and `Type.GetProperties()`) explicitly state that order is not guaranteed. So, I have changed the code to sort by `MemberInfo.MetadataToken` as advised in [this SO article](https://stackoverflow.com/questions/8067493/if-getfields-doesnt-guarantee-order-how-does-layoutkind-sequential-work).

Note that no unit tests have been added to cover this as it proved impossible to find a circumstance which reliably failed with the old code. It seems the order `GetFields` uses is usually declaration order but might change under certain runtime conditions. I have verified though that my change here fixed the issue I was seeing in my code.

### DIFF
--- a/csharp/RowOriented/ParquetFile.cs
+++ b/csharp/RowOriented/ParquetFile.cs
@@ -233,7 +233,23 @@ namespace ParquetSharp.RowOriented
                 list.Add((property.Name, mappedColumn, property.PropertyType, property));
             }
 
-            return list.ToArray();
+            // The order in which fields are processed is important given that when a tuple type is used in
+            // ParquetFile.CreateRowWriter<TTuple>() with an array of column names it is expected that
+            // the resulting parquet file correctly maps the name to the appropriate column type.
+            //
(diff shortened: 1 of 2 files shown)


GPSnoopy (approved):

jhickson:
I'm not sure why the MacOS build is failing - I haven't touched the native code.

GPSnoopy:
macOS failure is likely unrelated. Leave it with me.

GPSnoopy (approved):

PR #159, opened 2020-11-18
Writing decimals fails when column names provided

jhickson:
Writing a decimal column was fine unless you provide custom column names. When custom column names were provided, writing failed when copying the `Column` object to set the name because the copy failed to provide the `LogicalTypeOverride` and the `Length`. This fixes that and adds test coverage.

### DIFF
--- a/csharp.test/TestRowOrientedParquetFile.cs
+++ b/csharp.test/TestRowOrientedParquetFile.cs
@@ -148,7 +148,7 @@ namespace ParquetSharp.Test
 
         private static void TestRoundtrip<TTuple>(TTuple[] rows)
         {
-            RoundTripAndCompare(rows, rows);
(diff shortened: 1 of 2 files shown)


jhickson:
@GPSnoopy So if/when I make changes in the future I should avoid defaults in parameters then?

GPSnoopy:
@jhickson It's just a personal preference. I usually find it more readable to be explicit like in this case, especially for private method that don't have to keep the API compatible. I hope you don't mind that I committed directly onto your branch.

Will merge as soon as the tests pass.

GPSnoopy (approved):

GPSnoopy (approved):

jhickson:
@GPSnoopy I don't mind at all and I agree it can make the code clearer.

PR #162, opened 2020-12-03
Experimental AVX2 DateTime conversions.

GPSnoopy:
Target solely .NET Standard 2.0 and .NET Core 3.1.


GPSnoopy:
It does go slightly faster in some of the benchmarks, but not worth the AVX2 CPU downclock. Shelving the PR indefinitely.

PR #169, opened 2021-03-15
Setup the correct and fixed version of dotnet for each job, otherwise we are at the mercy of the cloud image.

GPSnoopy:


GPSnoopy:
Closed. Not the right approach as this is not how `dotnet-setup` works (yet?)

PR #170, opened 2021-03-15
Fix CI with .NET >= 5.0.200

jgiannuzzi:
This PR fixes recent CI woes since GitHub started deploying .NET 5.0.200 as the default SDK on their GitHub Actions runners (see failure examples [here](https://github.com/G-Research/ParquetSharp/pull/168/checks?check_run_id=2098347678) and [here](https://github.com/G-Research/ParquetSharp/pull/167/checks?check_run_id=2098093685)).

The behaviour of the `test-nuget` job is changed as follows: NuGet sources are specified in the right order in the top-level config and the implicit restore done during the package add will then use it instead of a unique source specified on the command line.

Before .NET 5.0.200, the implicit restore was also considering the nuget.org source, so this problem never occured.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,10 +170,14 @@ jobs:
         choco install nuget.commandline
         nuget add -source local nuget/ParquetSharp.${{ steps.get-version.outputs.version }}.nupkg
     - name: Add local NuGet feed
-      run: dotnet nuget add source -n local $PWD/local
+      run: |
+        dotnet nuget add source -n local $PWD/local
+        # Removing and adding back nuget.org ensures that our local source will be used first
+        dotnet nuget remove source nuget.org
+        dotnet nuget add source -n nuget.org https://api.nuget.org/v3/index.json
     - name: Change test project references to use local NuGet package
       run: |
         dotnet remove csharp.test reference csharp/ParquetSharp.csproj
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -170,10 +170,13 @@ jobs:
         choco install nuget.commandline
         nuget add -source local nuget/ParquetSharp.${{ steps.get-version.outputs.version }}.nupkg
     - name: Add local NuGet feed
-      run: dotnet nuget add source -n local $PWD/local
+      run: |
+        dotnet nuget add source -n local $PWD/local
+        dotnet nuget remove source nuget.org
```
GPSnoopy, line 175: Could use a comment, esp. why are we removing nuget.org and then adding it back again? Did you get this from a stackoverflow page? Maybe link it?

GPSnoopy:
Going to merge this is since that fixes the build and I can merge the other (unrelated) PRs that were blocked by this. Further improvements to be done as part of a separate PR (e.g. #171).

GPSnoopy (approved):

.github/workflows/ci.yml
jgiannuzzi: Good point, I have added a comment about it, though I'm not entirely sure it's actually doing anything (see https://github.com/NuGet/Home/issues/3676 for some context).

PR #174, opened 2021-03-17
Test dotnet 5

jgiannuzzi:
This PR modifies the test project to allow running on the .NET 5.0 runtime ~~when `-p:Channel=experimental` is specified~~. This requires building with the .NET 5.0 SDK, which might have an impact on the code we generate, though that should not be the case. It also requires contributors to have it installed to be able to build the tests.

This PR also modifies the CI workflow to run the tests on .NET 5.0 in parallel to .NET Core 3.1 and .NET Framework 4.6.1.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,8 @@ on:
 
 env:
   DOTNET_NOLOGO: true
-  dotnet_version: '3.1.407'
+  dotnet_3_version: '3.1.407'
+  dotnet_5_version: '5.0.201'
 
 jobs:
 
@@ -25,6 +26,7 @@ jobs:
       matrix:
         os: [ubuntu-18.04, macos-latest, windows-latest]
       fail-fast: false
+    name: Build native library (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     steps:
     - name: Checkout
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -180,17 +188,40 @@ jobs:
...
+            framework=netcoreapp3.1
+            ;;
+          4)
+            framework=net461
+            # We build for .NET framework with .NET Core 3.1 SDK
+            major=3
+            ;;
+          5)
+            framework=net5.0
+            options='-p:Channel=experimental'
```
GPSnoopy, line 207: I would not have thought that net5 was still considered experimental since it had now been released. Is this option really necessary?

jgiannuzzi: I have added that flag in the project file to avoid forcing users to install .NET 5.0 to build and run the tests. Happy to remove it if we can assume every contributors will have it installed.

GPSnoopy: Happy to keep it, but then needs a comment to explain why it's there :-)

jgiannuzzi: Makes sense, will add it once I get the chance 👌

jgiannuzzi: I figured out a way to automatically select the target frameworks based on runtime availability. See commit 5d35c21d196763905d6a4dc9814241124268a1b3.

jgiannuzzi: As discussed offline, I changed it once more to statically include .NET 5.0 as a target framework for the tests.

GPSnoopy (approved):

PR #177, opened 2021-04-08
Expose factories taking Reader/WriterProperties for row-oriented api

m00ngoose:


GPSnoopy:
From a first quick eyeballing, looks good to me.

Maybe some unit tests that proves that the reader/writer properties are passed down correctly? Doesn't need to be complex (I have done stupid mistakes in the past where half the arguments for the row-oriented API were not passed down correctly).

m00ngoose:
> From a first quick eyeballing, looks good to me.
> 
> Maybe some unit tests that proves that the reader/writer properties are passed down correctly? Doesn't need to be complex (I have done stupid mistakes in the past where half the arguments for the row-oriented API were not passed down correctly).

One way to test that both sets of properties propagate correctly would be something using encryption and decryption, throwing iff either is used without the other? Bit painful writing these without Rider :(

GPSnoopy:
For writers, we ought to expose the `WriterProperties`. For some reason, this bit of the API has not been exposed to C# yet. See #178. The reader properties are not exposed as such, so that part is a bit more complicated. Leave it to me.

GPSnoopy (approved):

Issue #178, opened 2021-04-08
Expose missing ParquetFileWriter properties


GPSnoopy:
WriterProperties, Schema, etc.

This would make unit testing easier in some cases. E.g. #177.

PR #180, opened 2021-04-09
Expose WriterProperties, SchemaDescriptor from ParquetFileWriter

m00ngoose:
https://github.com/G-Research/ParquetSharp/issues/178#issue-853647605


GPSnoopy:
Feels like we went through the same ordeal ;-)
See #179

PR #185, opened 2021-04-14
Add and expose LogicalReadConverterFactory to allow user reading custom types.

GPSnoopy:
## Goal

The aim of this PR is to expose the logic that maps the C# types (called "logical system types" by ParquetSharp, as per Parquet's LogicalType) to the actual Parquet physical types, as well as the converters that are associated with them.

This means that:
- a user can potentially read/write any type they want, as long as they provide a viable mapping,
- a user can override the default ParquetSharp mapping and change how existing C# types are handled.

## API

The API at the core of this PR is encompassed by `LogicalTypeFactory`, `LogicalReadConverterFactory` and `LogicalWriteConverterFactory`.

Whenever the user uses a custom type to read or write values to a Parquet file, a `LogicalRead/WriteConverterFactory` needs to be provided. This converter factory tells to the `LogicalColumnReader/Writer` how to convert the user custom type into a physical type that is understood by Parquet.

On top of that, if the custom type is used for creating the schema (when writing), or if accessing a `LogicalColumnReader/Writer` without explicitly overriding the element type (e.g. `columnWriter.LogicalReaderOverride<CustomType>()`), then a `LogicalTypeFactory` is needed in order to establish the proper logical type mapping.

In other words, the `LogicalTypeFactory` is required if the user provides a `Column` class with a custom type (writer only, the factory is needed to know the physical parquet type) or gets the `LogicalColumnReader/Writer` via the non type-overriding methods (in which case the factory is needed to know the full type of the logical column reader/writer). The corresponding converter factory is always needed.

## Refactoring

A good chunk of the PRs is just moving the existing type mapping from `Column` & `ColumnDescriptor` to `LogicalTypeFactory`, as well as exposing the read/write converters as public API (I've added a couple of utility methods to make the code less verbose and nicer to use, since it's all public now).

The rest is plumbing the factories throughout the code and ton-loads of unit tests.

## TODOs

- **Create tickets for the following, so we don't forget**
- Row-oriented API support (Separate PR)
- Kvp metadata avail for factories (aim to consume Arrows/Pandas metadata for things like timezones. (Separate PR or up to the user factory, TBD)


csharp.test/TestLogicalTypeFactory.cs
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        /// <summary>
+        /// A read converter factory that supports our custom type.
+        /// </summary>
+        private sealed class ReadConverterFactory : LogicalReadConverterFactory
+        {
+            public override Delegate GetDirectReader<TLogical, TPhysical>()
+            {
+                // Optional: the following is an optimisation and not stricly needed (but helps with speed).
```
m00ngoose, line 154: +nice

csharp/ColumnReader.cs
```diff
@@ -67,12 +74,17 @@ public void Dispose()
...
+            return LogicalColumnReader.Create(this, bufferLength, elementTypeOverride: null);
         }
 
         public LogicalColumnReader<TElement> LogicalReader<TElement>(int bufferLength = 4 * 1024)
         {
-            return LogicalColumnReader.Create<TElement>(this, bufferLength);
+            return LogicalColumnReader.Create<TElement>(this, bufferLength, elementTypeOverride: null);
+        }
+
+        public LogicalColumnReader<TElement> LogicalReaderOverride<TElement>(int bufferLength = 4 * 1024)
```
adamreeve, line 85: Would it be possible to avoid the need for the extra "Override" method and `elementTypeOverride` parameter? If the typed `LogicalReader` method receives a type parameter that's different to the default type provided by the LogicalTypeFactory, but is compatible with the `LogicalReadConverterFactory`, then it should be able to handle this, which would simplify the public API a bit.

csharp.test/TestColumn.cs
```diff
@@ -16,7 +16,7 @@ public static void TestPrimitives()
             {
                 Console.WriteLine("Testing primitive type {0}", expected.Type);
 
-                Assert.True(Column.IsSupported(expected.Type));
+                //Assert.True(Column.IsSupported(expected.Type));
```
adamreeve, line 19: Tidy up needed here?

csharp/LogicalReadConverterFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// Extendable class that handles the mapping between a Parquet physical type and the C# logical type when reading values.
+    /// </summary>
+    public class LogicalReadConverterFactory
+    {
+        /// <summary>
+        /// Return a reader delegate if a TPhysical column reader can directly write into a TLogical span (e.g. float to float, int to uint, etc).
+        /// Otherwise return null. This is an optimisation to avoid needless memory copies between buffers (i.e. otherwise we have to use the
+        /// identity converter).
+        /// </summary>
+        public virtual Delegate? GetDirectReader<TLogical, TPhysical>()
```
adamreeve, line 16: Why use an untyped Delegate as the return type here and elsewhere in the public API? It's not very obvious for users inheriting from this class what exactly they should be returning without digging into the code.

GPSnoopy: Surprisingly makes the code a lot clearer elsewhere. See the diffs between `LogicalRead.cs` (same for write):

`return (DirectReader) (Delegate) (LogicalRead<TPhysical, TPhysical>.DirectReader) Read;` 
vs
`return LogicalRead.GetDirectReader<TPhysical, TPhysical>();`

The problem is due to how C# generics are handled, you always need to cast to `Delegate` before you can recast up. The new code now only does the re-cast to the type-safe delegate in one place, and the factories are okay to return `Delegate` for code simplicity. Old code had to do that cast everywhere otherwise.

csharp.test/TestColumn.cs
GPSnoopy: Fixed

csharp/ColumnReader.cs
adamreeve: 👍 makes sense

csharp/LogicalReadConverterFactory.cs
adamreeve: OK that makes sense, it would be good to have the expected types documented in the XML docs for LogicalReadConverterFactory and LogicalWriteConverterFactory though to make this a bit clearer for users (eg. maybe just a reference to the typed delegates).

GPSnoopy: Done

adamreeve (approved):

rgr21 (approved):

csharp/ColumnDescriptor.cs
```diff
@@ -59,9 +64,9 @@ public TReturn Apply<TReturn>(IColumnDescriptorVisitor<TReturn> visitor)
         /// LogicalType is the most nested logical type (e.g. string).
         /// ElementType is the type represented by the column (e.g. string[][][]).
         /// </summary>
-        public (Type physicalType, Type logicalType, Type elementType) GetSystemTypes()
+        public (Type physicalType, Type logicalType, Type elementType) GetSystemTypes(LogicalTypeFactory typeFactory, Type? columnLogicalTypeOverride)
```
m00ngoose, line 67: could we have a defaulted overload or default values on this? I guess most of the time you'll want LogicalTypeFactory.Default and null.

m00ngoose: Given that it's part of the public api I mean

m00ngoose:
This PR changed the public api in at least two places (CreateSchemaNode, GetSystemTypes).

Is it worth having a changelog / migration guide? Making sure there are defaulted versions of the old api?

GPSnoopy:
@m00ngoose Not a bad comment, we should probably leave the previous overload with default values. I'll add that to the ticket.

Issue #186, opened 2021-04-17
Recursive row reading/writing


m00ngoose:
Copied from #177 

```
public readonly struct UserId {
    public readonly int UserId;
    public UserId(int userId) => UserId = userId;
}

public class User {
    public readonly UserId UserId;
    public readonly string Name;
    public readonly int Age;
    public User(UserId userId, string name, int age) {
        UserId = userId;
        Name = name;
        Age = age;
    }
}
```

Another way of being able to roundtrip this using the row-oriented api, is by automatically flattening nested classes.

GPSnoopy:
How is this different from the official Parquet Format nested structure representation?

Things get rather messy when dealing with nullable types when they're nested. Have a look at the nested array logic in the logical reader and writer implementations.

m00ngoose:
I believe this is orthogonal from Parquet-native nested structure representation, because I'm traversing the nested C#-representation to create a flat list of Parquet columns.

I'm not sure what difficulties you envisage re. nullable types - can you give an example class/struct hierarchy?

I claim this approach would let strictly more types roundtrip through the row-oriented api. The only downside I see is losing the ability to do a more "faithful" native-nested-representation in future without breaking backwards-compatibility.

GPSnoopy:
> I'm traversing the nested C#-representation to create a flat list of Parquet columns

This is exactly that. That's why Parquet provides definition and repetition levels, which you can access using the low-level `ColumnWriter`.

For the moment I chose to not include this in ParquetSharp given that Apache Arrow itself does not support it (yet?).

Let's have a chat at some point in the office.

GPSnoopy:
Won't fix. Out of scope for ParquetSharp for the moment.

GPSnoopy:
IMHO, supporting such a feature is quite a lot of work as it requires the following:
- Reflection to traverse the nested structures.
- Reflection attributes to control serialisation behaviour (each user of the row-oriented API has a different opinion on how it should behave by default).
- Full support of repetition/definition-level to know when a field/nested-field is `null`.

This would lead to a wide interface in ParquetSharp to support all possible use cases; hinting that this is not the right place or the right responsibility to consolidate such a set of feature.

On the other hand, flattening a set of nested structs to either a flat structure (for the row-oriented API) or to an array of columns (for the low-level API) is a cheap coding operation user-side (if the structure is known in advance).

PR #187, opened 2021-04-17
Recursive row (de)serialization

m00ngoose:
POC for #186 

Code is very messy, but, unit tests pass.

TODO: 
1. Fail gracefully for recursive types.
2. Column names? Should perhaps be eg. "Outer.Inner.Field" rather than just "Field"

Q: Is this worth pursuing further?


m00ngoose:
Agreed not to go down this route.

PR #193, opened 2021-05-11
Make Date : IComparable<Date>, DateTimeNanos : IComparable<DateTimeNanos>

m00ngoose:
Was trying to sort some rows generically, it failed because Date is not IComparable.


GPSnoopy (approved):
Approved. But I want PR #194 to be merged in first to avoid potential problems.

PR #196, opened 2021-05-18
add code format check

ljubon:
Force formatting code with predefined rules in `.editorconfig` for each `push` and `pr`
The goal is to reformat all current code, after which we should enable the `dotnet-format` job only in `pr` as required job to finish with code `0`.
If some formating is found in new code, the job will return `exit code 2` and stop further jobs until no formating is needed.

**added**:
- `dotnet-format` tool, version `5.1.225507` (latest stable)
- make sure the 'Check format' job only runs once
- use rules from `.editorconfig` in the root of the project
- current reformatting is minimal
  - Fix whitespace formatting
  - Fix final newline
  - Fix end of line marked
  - Fix imports ordering

Knowledge base for rules; https://www.jetbrains.com/help/resharper/EditorConfig_CSHARP_LineBreaksPageSchema.html


.github/workflows/ci.yml
```diff
@@ -14,6 +14,22 @@ jobs:
   # Note that vcpkg dependencies takes the majority of the build time.
   # We cache them using GitHub Actions cache and export, making the scripts below a bit more complex.
 
+  # Format file with instruction in ~/.editorconfig file
```
jgiannuzzi, line 17: We want to enforce the consistency of code formatting, not to format — Could you please update the comment to reflect that?

.github/workflows/ci.yml
```diff
@@ -14,6 +14,22 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup .NET Core SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: ${{ env.dotnet_3_version }}
+      - name: Restore tools
+        run: dotnet tool restore
+      - name: Check format
+        run: dotnet format "." -f --check
```
jgiannuzzi, line 31: I think we should not use the `-f` option, but instead directly the project files/dirs. If we add a solution file referencing our 3 projects, then `dotnet format --check` will work as-is.

@GPSnoopy what are your thoughts on adding a solution file?

.config/dotnet-tools.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "dotnet-format": {
+      "version": "4.1.131201",
```
jgiannuzzi, line 6: Let's use the latest release (5.1.225507), unless there is a good reason not to?

.editorconfig
```diff
@@ -0,0 +1,207 @@
...
+
+# New line preferences
+end_of_line = crlf
+insert_final_newline = true
+
+#### .NET Coding Conventions ####
+
+# Organize usings
+dotnet_separate_import_directive_groups = false
+dotnet_sort_system_directives_first = false
```
jgiannuzzi, line 23: I think `dotnet_sort_system_directives_first` should be `true` to better reflect the current project style.

@GPSnoopy do you confirm?

.editorconfig
```diff
@@ -0,0 +1,207 @@
...
+# Indentation preferences
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = true
+csharp_indent_labels = one_less_than_current
+csharp_indent_switch_labels = true
+
+# Space preferences
+csharp_space_after_cast = false
```
jgiannuzzi, line 140: @GPSnoopy I have seen both styles used in the project for `csharp_space_after_cast`. Which one do you feel is the 'right' one?

README.md
```diff
@@ -144,7 +144,8 @@ We have had to write our own `FindPackage` macros for most of the dependencies t
 
 ## Contributing
 
-We welcome new contributors! We will happily receive PRs for bug fixes or small changes. If you're contemplating something larger please get in touch first by opening a GitHub Issue describing the problem and how you propose to solve it.
+We welcome new contributors! We will happily receive PRs for bug fixes or small changes. If you're contemplating something larger please get in touch first by opening a GitHub Issue describing the problem and how you propose to solve it. 
+Once PR is oppened, make sure that `check-format` job is passing before checking `PR` as ready for review.
```
jgiannuzzi, line 148: Nice one! I would probably rephrase to get contributors to install and run `dotnet-format` locally before creating a PR. And mention of couple of popular IDEs that already support `.editorconfig` and will just do the right thing for them when they save the files.

jgiannuzzi (commented):
Thanks a lot for your PR, @ljubon!

I made a couple of comments below, but this is overall solid work 👍

Could you please format the files with `dotnet format` in this PR too? This will help with discussing and fine-tuning the formatting rules.

.github/workflows/ci.yml
jgiannuzzi: We could use this comment taken verbatim from [Consul.NET](https://github.com/G-Research/consuldotnet):
> Enforces the consistency of code formatting using `.editorconfig` and the `dotnet-format` tool.

.github/workflows/ci.yml
jgiannuzzi: Indeed this works fine for the CI, but for contributors this will start formatting 3rd party dependencies found under the `build` directory. Adding a solution file would solve that and make the command easier. I suggest you go ahead and add a solution file, unless @GPSnoopy has a strong opinion against.

.config/dotnet-tools.json
ljubon: fixed 👍

.github/workflows/ci.yml
ljubon: It makes sense for me, pushed new version, check it out @GPSnoopy t @jgiannuzzi  👀

.editorconfig
ljubon: pushed with `dotnet_sort_system_directives_first = true`
let me know how to proceed 🕺

README.md
ljubon: I was in the loop for some time about some words, but here is the result!

.editorconfig
ljubon: I'll wait for the final `.editoconfig` until I format files locally and push it as a separate commit
if you have any suggestions, I'm open to work on it!

.github/workflows/ci.yml
jgiannuzzi: It looks like the new version somehow didn't get pushed 😅

.editorconfig
jgiannuzzi: I think you can already push the formatted files now, so we can discuss and possibly fine-tune the formatting rules over here.

.config/dotnet-tools.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "dotnet-format": {
+      "version": "5.1.225507",
```
ljubon, line 6: @jgiannuzzi per dotnet-format repo, this one is the latest one https://github.com/dotnet/format/releases/tag/v5.1.225507

.github/workflows/ci.yml
ljubon: hey @jgiannuzzi, version is noted in .config\dotnet-tools.json 
since we are doing `dotnet tool restore` it's picking version set there

jgiannuzzi: I meant the new version of your PR adding the solution file and changing the command to check the format 😄

.editorconfig
GPSnoopy: `csharp_space_after_cast` definitely true. VS + Resharper are sometimes inconsistent. 

Not keen on a solution file since it's currently auto-generated by CMake

.github/workflows/ci.yml
jgiannuzzi: As per @GPSnoopy's [comment](https://github.com/G-Research/ParquetSharp/pull/196#discussion_r636976230) below, he is not keen on a solution file since it's currently auto-generated by CMake.
Do we want to go for something like `for project in */*.csproj; do dotnet format --check $project; done` then?

.editorconfig
ljubon: hi @GPSnoopy @jgiannuzzi , thank you for your help
FYI I pushed already new commit with all files formated +  `csharp_space_after_cast = true`

.github/workflows/ci.yml
ljubon: also, we can do with multiple commands, since it will be cleaner (readable) and will not affect the time of the job itself much

`      - name: Check format - ParquetSharp
        run: dotnet format "./csharp/ParquetSharp.csproj"
      - name: Check format - Benchmark
        run: dotnet format "./csharp/ParquetSharp.Benchmark.csproj"
      - name: Check format - Test
        run: dotnet format "./csharp/ParquetSharp.Test.csproj"
`

jgiannuzzi: works for me 👍 — just make sure to add `--check` 🤓

jgiannuzzi:
@GPSnoopy What are your thoughts on the formatting rules? I think they are pretty close to the existing style, except maybe for `{ 1, 2, 3 }` vs `{1, 2, 3}` (which I'm not sure can be configured in `.editorconfig`).

GPSnoopy:
I prefer `{1, 2, 3}` if possible

GPSnoopy:
I'll have a look at the rules and full formatting early next week

.github/workflows/ci.yml
ljubon: added all paths and `--check` 🤓

.github/workflows/ci.yml
```diff
@@ -14,6 +14,26 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup .NET Core SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: ${{ env.dotnet_3_version }}
+      - name: Restore tools
+        run: dotnet tool restore
+      - name: Check format - ParquetSharp.csproj
```
jgiannuzzi, line 30: nitpick: I would say 'Library' instead of 'ParquetSharp.csproj'

jgiannuzzi:
> I prefer `{1, 2, 3}` if possible

Unfortunately this does not seem to be supported by `dotnet-format`, which only implements the official [C# formatting rules](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#c-formatting-rules).

ReSharper supports it as [`resharper_csharp_space_within_single_line_array_initializer_braces`](https://www.jetbrains.com/help/resharper/EditorConfig_CSHARP_SpacesPageSchema.html#resharper_csharp_space_within_single_line_array_initializer_braces), so maybe we set that one to `true` to ensure the same experience between ReSharper and `dotnet-format`.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,26 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup .NET Core SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: ${{ env.dotnet_3_version }}
+      - name: Restore tool
+        run: dotnet tool restore
+      - name: Check format - Library
+        run: dotnet format "./csharp/ParquetSharp.csproj" --check
```
jgiannuzzi, line 31: We could simplify those by pointing to the project directory, i.e. `dotnet format csharp --check`

.editorconfig
```diff
@@ -0,0 +1,207 @@
...
+#### Core EditorConfig Options ####
+
+# Indentation and spacing
+indent_size = 4
+indent_style = space
+tab_width = 4
+trim_trailing_whitespace = true
+
+# New line preferences
+end_of_line = crlf
```
jgiannuzzi, line 16: This might be a bad interaction with git, but your latest commits all seem to end with LF, not CRLF. Maybe we should change this pref to `lf` to avoid bad interactions with git? (This is just a theory I did not validate - I might be wrong)

.github/workflows/ci.yml
ljubon: nicee! I missed that, pushing it now

.editorconfig
ljubon: I checked locally, all files `*.cs` are set as `CRLF` - that's matching `.editorconfig` so it should work whatever we set here
Logically, C# devs should work from Win platform mostly, so I would leave like this `end_of_line = crlf`

jgiannuzzi: Hmm, the problem is that the CI runs on Linux and thus fails. Also there are devs working on macOS (like me) or Linux that then get all files reformatted when running `dotnet format`. I think we should get rid of the `end_of_line` directive to avoid those issues, as AFAIU git will checkout files with the platform's native line endings anyway.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,26 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup .NET Core SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: ${{ env.dotnet_3_version }}
+      - name: Restore tool
+        run: dotnet tool restore
+      - name: Check format - Library
+        run: dotnet format "./csharp/" --check
```
jgiannuzzi, line 31: nitpick: any reason to have `"./csharp/"` instead of just `csharp`

.github/workflows/ci.yml
ljubon: i'll resolve this comments, as new change is using `jb cleanupcode` command so we can discuss that separatly

ljubon:
> > I prefer `{1, 2, 3}` if possible
> 
> Unfortunately this does not seem to be supported by `dotnet-format`, which only implements the official [C# formatting rules](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#c-formatting-rules).
> 
> ReSharper supports it as [`resharper_csharp_space_within_single_line_array_initializer_braces`](https://www.jetbrains.com/help/resharper/EditorConfig_CSHARP_SpacesPageSchema.html#resharper_csharp_space_within_single_line_array_initializer_braces), so maybe we set that one to `true` to ensure the same experience between ReSharper and `dotnet-format`.

Hi both 😃 @GPSnoopy @jgiannuzzi 
first of all, sorry for spamming with each new push, I couldn't find how to disable Actions for this PRs 😕 
As you can see with latest commit, I've switched using `jb cleanupcode` which can give us more options and rules to apply
At the moment only change is enforcing now  [`resharper_csharp_space_within_single_line_array_initializer_braces`](https://www.jetbrains.com/help/resharper/EditorConfig_CSHARP_SpacesPageSchema.html#resharper_csharp_space_within_single_line_array_initializer_braces) as it's requested.

Another thing, which somehow I couldn't do is restore `jb` as a tool and avoid any installation, any thought about that?

jgiannuzzi:
Great stuff @ljubon!

It looks like `jb codecleanup` is unfolding a few expressions in multiple lines. This might be due to an implicit line length limit. Could you please try to change that so it matches the current style?

ljubon:
> Great stuff @ljubon!
> 
> It looks like `jb codecleanup` is unfolding a few expressions in multiple lines. This might be due to an implicit line length limit. Could you please try to change that so it matches the current style?

Thanks @jgiannuzzi! Could you point out some examples so I can try find matches with some rules?

csharp.benchmark/Check.cs
```diff
@@ -21,9 +21,10 @@ public static void ArraysAreEqual<T>(T[] expected, T[] result)
             {
                 if (!comparer.Equals(expected[i], result[i]))
                 {
-                    throw new InvalidDataException($"expected value {expected[i]} != result value {result[i]} at index {i}");
+                    throw new InvalidDataException(
+                        $"expected value {expected[i]} != result value {result[i]} at index {i}");
```
jgiannuzzi, line 25: Here is an example of a line that got split into two lines, probably because it was 'too long' according to the current settings (just a guess)

GPSnoopy:
In general I don't like auto-formatters splitting long lines (having a 32" 4K screen might be related). Happy if it leaves long lines alone, and leave it to the maintainers.

csharp.benchmark/Check.cs
ljubon: finally got it! please, when you can enable the latest CI
FYI I found out that there is no way to generate `ParquetSharp.DotSettings` via some CLI. The only way to do it is to import the project in ReSharper IDE and check which options you would like to have (Code Style -> C#). In docc they are referencing this concept as [Layers](https://www.jetbrains.com/help/rider/Sharing_Configuration_Options.html). This file was generated as part of [Solution-Team shared](https://www.jetbrains.com/help/rider/Sharing_Configuration_Options.html#solution-team-shared-layer) layer.
Let me know if you found any other instances, of irregularity since with example I can easily find and match which rule needs to be changed. 
Thank you guys! @GPSnoopy @jgiannuzzi 🍻

csharp.benchmark/FloatTimeSeriesBase.cs
```diff
@@ -7,33 +7,21 @@ public abstract class FloatTimeSeriesBase
     {
         protected static Column[] CreateFloatColumns()
         {
-            return new Column[]
-            {
-                new Column<DateTime>("DateTime", LogicalType.Timestamp(true, TimeUnit.Millis)),
-                new Column<int>("ObjectId"),
-                new Column<float>("Value")
-            };
+            return new Column[] {new Column<DateTime>("DateTime", LogicalType.Timestamp(true, TimeUnit.Millis)), new Column<int>("ObjectId"), new Column<float>("Value")};
```
jgiannuzzi, line 10: The previous style (one item per line) was more readable. Not sure why ReSharper felt like reformatting it on one line 😒
@ljubon can you please try to find a setting to avoid this reformatting?

csharp.benchmark/FloatTimeSeriesBase.cs
```diff
@@ -7,33 +7,21 @@ public abstract class FloatTimeSeriesBase
...
 
         protected static (DateTime[] dates, int[] objectIds, float[][] values, int numRows) CreateFloatDataFrame(int numDates)
         {
             var rand = new Random(123);
 
-            var dates = Enumerable.Range(0, numDates)
-                .Select(i => new DateTime(2001, 01, 01) + TimeSpan.FromHours(i))
-                .Where(d => d.DayOfWeek != DayOfWeek.Saturday && d.DayOfWeek != DayOfWeek.Sunday)
-                .ToArray();
+            var dates = Enumerable.Range(0, numDates).Select(i => new DateTime(2001, 01, 01) + TimeSpan.FromHours(i)).Where(d => d.DayOfWeek != DayOfWeek.Saturday && d.DayOfWeek != DayOfWeek.Sunday).ToArray();
```
jgiannuzzi, line 17: Same remark as above. The previous style was more readable (one chained method call per line).

csharp.benchmark/Program.cs
```diff
@@ -44,9 +34,7 @@ public static int Main()
                 }
 
                 return 0;
-            }
-
-            catch (Exception exception)
+            } catch (Exception exception)
```
jgiannuzzi, line 37: Same remark as above. The previous style was more readable (ending brace and catch on a separate line).

csharp.test/PhysicalValueGetter.cs
```diff
@@ -10,8 +10,7 @@ public PhysicalValueGetter(long numValues)
             _numValues = numValues;
         }
 
-        public (Array values, short[] definitionLevels, short[] repetitionLevels) OnColumnReader<TValue>(ColumnReader<TValue> columnReader) 
-            where TValue : unmanaged
+        public (Array values, short[] definitionLevels, short[] repetitionLevels) OnColumnReader<TValue>(ColumnReader<TValue> columnReader) where TValue : unmanaged
```
jgiannuzzi, line 13: Same remark as above. The previous style was more readable (`where` on a separate line).

csharp.test/PhysicalValueGetter.cs
```diff
@@ -21,17 +20,15 @@ public PhysicalValueGetter(long numValues)
 
             while (columnReader.HasNext)
             {
-                var levelsRead = columnReader.ReadBatch(
-                    _numValues - totalLevels, defLevels.AsSpan(totalLevels), repLevels.AsSpan(totalLevels), values.AsSpan(totalValues), 
-                    out var valuesRead);
+                var levelsRead = columnReader.ReadBatch(_numValues - totalLevels, defLevels.AsSpan(totalLevels), repLevels.AsSpan(totalLevels), values.AsSpan(totalValues), out var valuesRead);
```
jgiannuzzi, line 23: Same remark as above. The previous style was more readable (`in` and `out` arguments on different lines).

jgiannuzzi (commented):
Great effort, @ljubon!

I have highlighted a few unwelcome changes to the formatting below.
As a rule of thumb, it would be great if there was a way to tell `codecleanup` to _not_ be as agressive with the reformatting, especially when the developer has split multiples arguments/items/method calls between lines.

Compared to `dotnet format`, it seems that we gained the ability to use `csharp_space_within_single_line_array_initializer_braces = true`, but at a detriment of more agressive reformatting (so far). Let's keep exploring `codecleanup` and hopefully solve those unwelcome changes, whilst keeping in mind that `dotnet format` comes pretty close to what we need.

csharp.benchmark/FloatTimeSeriesBase.cs
ljubon: Thanks for pointing this out! It's a great example to test these tools further.
I'll check with ReSharper how to keep current re-formating with addition for this rule (not sure how to name it 🤔 )
Also, I would agree with choosing `dotnet-format ` as the final tool too, but I wanted to challenge it with `cleanupcode` where we have options in our hand, but as they say sometimes _less is more_ :)

ljubon: with a bit of tuning, it seems that all examples below (and others too) are reverted and added to `ParquetSharp.DotSettings` 👍 @jgiannuzzi @GPSnoopy  let me know if you find anything similar 👀

GPSnoopy:
Hi @ljubon,

Impressive work. 

I've added a few comments. The `finally`/`else` on the same line as the closing `}` does look very odd to me, same with the first argument on the same line as the method declaration while the other arguments are left untouched. If we can fix those I'd be very happy.

Leaving attributes line-return untouched would be nice to have.

Do note that I'm using ReSharper to develop ParquetSharp. I've not changed many default options either (mainly I've disabled the "wrap long lines").

If it helps I can share my local ReSharper config.

ljubon:
> Hi @ljubon,
> 
> Impressive work.
> 
> I've added a few comments. The `finally`/`else` on the same line as the closing `}` does look very odd to me, same with the first argument on the same line as the method declaration while the other arguments are left untouched. If we can fix those I'd be very happy.
> 
> Leaving attributes line-return untouched would be nice to have.
> 
> Do note that I'm using ReSharper to develop ParquetSharp. I've not changed many default options either (mainly I've disabled the "wrap long lines").
> 
> If it helps I can share my local ReSharper config.

Big thanks @GPSnoopy 
Yeah! I would like to check your's ReSharper config, do note that this one I created like `team layer` as they described in docc. Not sure if yours is just for ParquetSharp or All DotNet/C# projects you have on your machine (if that's case then if you can please send both your $HOME config and ParquetSharp config 🙏 )

Also, for comments above, since I'm not in DotNet/C# world (yet 🤞 ) could you please point out examples like @jgiannuzzi did above? It would be highly appreciated! 🙌 
Thanks!

GPSnoopy:
Hi @ljubon, since ReSharper only exports the settings I've modified (but not the default settings), this is the only relevant line I have. Which is probably not very helpful.

```xml
<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_LINES/@EntryValue">False</s:Boolean>
```

ljubon:
> Hi @ljubon, since ReSharper only exports the settings I've modified (but not the default settings), this is the only relevant line I have. Which is probably not very helpful.
> 
> ```
> <s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_LINES/@EntryValue">False</s:Boolean>
> ```

Hey @GPSnoopy I think I got this `finally / else` formatting, please check when you can 🙏 
FYI maybe you can give me more hints from this link for some rules about [new-line formatting](https://docs.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/style-rules/formatting-rules?view=vs-2019#new-line-options) or some more examples like here https://github.com/RehanSaeed/EditorConfig/blob/main/.editorconfig

csharp.test/TestEncryption.cs
```diff
@@ -226,15 +226,14 @@ private static FileDecryptionProperties CreateDecryptAllSameKeyProperties()
...
             return builder
                 .KeyRetriever(new TestRetriever())
                 .Build();
         }
 
-        private static void AssertEncryptionRoundtrip(
-            Func<FileEncryptionProperties?> createFileEncryptionProperties,
-            Func<FileDecryptionProperties?> createFileDecryptionProperties, 
+        private static void AssertEncryptionRoundtrip(Func<FileEncryptionProperties?> createFileEncryptionProperties,
+            Func<FileDecryptionProperties?> createFileDecryptionProperties,
```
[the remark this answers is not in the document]
GPSnoopy: Yes, I already pointed this one. We definitely do not want that. Splitting args this way is definitely a no go.

csharp/LogicalType.cs
```diff
@@ -169,27 +168,37 @@ internal static LogicalType Create(IntPtr handle)
 
     public sealed class StringLogicalType : LogicalType
     {
-        internal StringLogicalType(IntPtr handle) : base(handle) { }
+        internal StringLogicalType(IntPtr handle) : base(handle)
+        {
+        }
```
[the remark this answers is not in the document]
GPSnoopy: Nice to have. Not a blocker though

csharp.test/TestRowOrientedParquetFile.cs
```diff
@@ -125,14 +124,15 @@ public static void TestWriterDoubleDispose()
         }
 
         [Test]
-        public static void TestCompressionArgument([Values(Compression.Uncompressed, Compression.Brotli)] Compression compression)
+        public static void TestCompressionArgument([Values(Compression.Uncompressed, Compression.Brotli)]
+            Compression compression)
```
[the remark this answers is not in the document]
GPSnoopy: Preferably not.  As pointed out earlier, the reformatting seems to want to put a line-break after the attribute here. But then does the exact opposite a few lines below (i.e. removes line-breaks after attributes on fields).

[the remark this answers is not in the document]
ljubon: I couldn't find any suitable pattern for this, as it's not affected by wrap_line nor number of parameters inside method
Would you say that this line 127 is more like [Arrangement_of_attributes](https://www.jetbrains.com/help/resharper/EditorConfig_CSHARP_LineBreaksPageSchema.html#Arrangement_of_attributes) or [Arrangement_of_method_signatures](https://www.jetbrains.com/help/resharper/EditorConfig_CSHARP_LineBreaksPageSchema.html#Arrangement_of_method_signatures) ?
I tested a few, but it's not affecting this line 128 🤔

csharp/LogicalType.cs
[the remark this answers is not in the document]
ljubon: fixed

.github/workflows/ci.yml
```diff
@@ -13,6 +13,30 @@ jobs:
...
+          dotnet tool restore
+          dotnet jb cleanupcode "csharp" "csharp.test" "csharp.benchmark" --profile="Built-in: Reformat Code" --settings="ParquetSharp.DotSettings" --dotnetcoresdk=${{ env.dotnet_5_version }} --verbosity=WARN
+          check=$(git diff --name-only | wc -l)
+          if [ $check -eq 0 ]
+          then
+            echo "Check passed"
+            exit 0
+          else
+            echo "Code format check - FAILED"
+            git diff --name-only
```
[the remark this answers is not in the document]
ljubon: Thanks for this hint, it's really good locking in job output! 🤞 
i couldn't make it work on my zsh with `$1` as arguments (to get file name) so I use it awk instead to grab it
`git diff --name-only | echo "::error file=$(awk '{print $1}') ::Code format check failed"`
it should work like charm on `bash`

csharp.test/TestRowOrientedParquetFile.cs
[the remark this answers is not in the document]
ljubon: fixed 🙏

GPSnoopy:
I think the latest patch is making things worse...

I'm trying `jb cleanup` locally with a fresh checkout and playing with it to find the right settings (I'm not far off I think), but for some reason it ignores my `DotSettings` file (it says it's using it, but not matter what settings I put in there they have no effect). @ljubon Am I missing something obvious?

`jb cleanupcode --profile="Built-in: Reformat Code" --settings="Formatting.DotSettings"  "csharp" "csharp.test" "csharp.benchmark"`

```xml
<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
	<s:Boolean X:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_LINES/@EntryValue">False</s:Boolean>
	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/LINE_FEED_AT_FILE_END/@EntryValue">True</s:Boolean>
	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/PLACE_ACCESSOR_ATTRIBUTE_ON_SAME_LINE_EX/@EntryValue">NEVER</s:String>
	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/PLACE_ACCESSORHOLDER_ATTRIBUTE_ON_SAME_LINE_EX/@EntryValue">NEVER</s:String>
	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/PLACE_FIELD_ATTRIBUTE_ON_SAME_LINE_EX/@EntryValue">NEVER</s:String>
</wpf:ResourceDictionary> 
```

GPSnoopy:
Never mind. There was a typo (`X:Key` instead of `x:Key`) which was causing it to silently fail...

GPSnoopy:
This is what I got when the command line on my local workstation:
https://github.com/G-Research/ParquetSharp/pull/201

This is pretty much what we want. Note that I've not put a `.editorconfig` file. I wanted to start from scratch,

ljubon:
files are now matching https://github.com/G-Research/ParquetSharp/pull/201  💯

jgiannuzzi:
One disadvantage I see with using `DotSettings` instead of `.editorconfig` is that it doesn't get picked up automatically by ReSharper/Rider and so contributors will have to manually run the reformatting command in many cases.

I have tested a few different things locally and realised that the `.editorconfig` file has to be present in each project directory because we don't have a solution file.

This is a work in progress, but placing the following `.editorconfig` in each project directory (`csharp`, `csharp.benchmark`, `csharp.test`) and calling `dotnet jb codecleanup csharp*` gives me a result that is pretty close to #201:
```ini
[*.{csproj,json,targets}]
indent_size = 2
resharper_xml_wrap_tags_and_pi = false

[*.cs]
csharp_new_line_before_members_in_object_initializers = false
csharp_preserve_single_line_blocks = true
resharper_blank_lines_after_block_statements = 0
resharper_blank_lines_around_auto_property = 0
resharper_blank_lines_around_single_line_type = 0
resharper_csharp_blank_lines_around_field = 0
resharper_csharp_insert_final_newline = true
resharper_csharp_wrap_lines = false
resharper_empty_block_style = multiline
resharper_keep_existing_embedded_block_arrangement = false
resharper_keep_existing_enum_arrangement = false
resharper_max_attribute_length_for_same_line = 70
resharper_place_accessorholder_attribute_on_same_line = false
resharper_place_expr_accessor_on_single_line = true
resharper_place_expr_method_on_single_line = true
resharper_place_expr_property_on_single_line = true
resharper_place_field_attribute_on_same_line = false
resharper_place_simple_embedded_statement_on_same_line = true
resharper_place_simple_initializer_on_single_line = false
resharper_remove_blank_lines_near_braces_in_code = false
resharper_remove_blank_lines_near_braces_in_declarations = false
resharper_wrap_array_initializer_style = chop_if_long
resharper_wrap_chained_binary_expressions = chop_if_long
resharper_wrap_object_and_collection_initializer_style = chop_always
```

I generated it with Rider's "Auto-Detect Formatting Rules" functionality, then tweaked it a bit further.

.github/workflows/ci.yml
```diff
@@ -13,6 +13,30 @@ jobs:
...
+          check=$(git diff --name-only | wc -l)
+          if [ $check -eq 0 ]
+          then
+            echo "Check passed"
+            exit 0
+          else
+            echo "Code format check - FAILED"
+            git diff --name-only | echo "::error file=$(awk '{print $1}') ::Code format check failed"
+            exit 1
+          fi
```
jgiannuzzi, line 39: This will not output the proper error. Instead there should be one error per badly-formatted file.
We can also drop the positive case and call `git diff` just once with something like this:

```suggestion
          files=($(git diff --name-only))
          if [ ${#files[@]} -gt 0 ]
          then
            for file in $files; do echo "::error file=$file::Code format check failed"; done
            exit 1
          fi
```

GPSnoopy:
How about putting the DotSettings file next to the SLN as part of the CMake SLN generation?

jgiannuzzi:
> How about putting the DotSettings file next to the SLN as part of the CMake SLN generation?

When and how exactly does that CMake SLN generation happen? I can't seem to get one on macOS and Linux :/

.github/workflows/ci.yml
ljubon: pure beauty 🙏 👍

GPSnoopy:
You need to call `cmake` with the right generator. To simply use the VS one, add `-G "Visual Studio 16 2019" -A "x64"`.

Also I was thinking maybe we want to add the formatting checks as part of a new YAML file? Doesn't have to be in `ci.yml` as it's completely separate.

Be careful when naively calling `jb cleanupcode` on the solution, as it will then also do the C++ files (which I guess we don't want at this point)

ljubon:
> One disadvantage I see with using `DotSettings` instead of `.editorconfig` is that it doesn't get picked up automatically by ReSharper/Rider and so contributors will have to manually run the reformatting command in many cases.
> 
> I have tested a few different things locally and realised that the `.editorconfig` file has to be present in each project directory because we don't have a solution file.
> 
> This is a work in progress, but placing the following `.editorconfig` in each project directory (`csharp`, `csharp.benchmark`, `csharp.test`) and calling `dotnet jb codecleanup csharp*` gives me a result that is pretty close to #201:
> 
> ```ini
> [*.{csproj,json,targets}]
> indent_size = 2
> resharper_xml_wrap_tags_and_pi = false
> 
> [*.cs]
> csharp_new_line_before_members_in_object_initializers = false
> csharp_preserve_single_line_blocks = true
> resharper_blank_lines_after_block_statements = 0
> resharper_blank_lines_around_auto_property = 0
> resharper_blank_lines_around_single_line_type = 0
> resharper_csharp_blank_lines_around_field = 0
> resharper_csharp_insert_final_newline = true
> resharper_csharp_wrap_lines = false
> resharper_empty_block_style = multiline
> resharper_keep_existing_embedded_block_arrangement = false
> resharper_keep_existing_enum_arrangement = false
> resharper_max_attribute_length_for_same_line = 70
> resharper_place_accessorholder_attribute_on_same_line = false
> resharper_place_expr_accessor_on_single_line = true
> resharper_place_expr_method_on_single_line = true
> resharper_place_expr_property_on_single_line = true
> resharper_place_field_attribute_on_same_line = false
> resharper_place_simple_embedded_statement_on_same_line = true
> resharper_place_simple_initializer_on_single_line = false
> resharper_remove_blank_lines_near_braces_in_code = false
> resharper_remove_blank_lines_near_braces_in_declarations = false
> resharper_wrap_array_initializer_style = chop_if_long
> resharper_wrap_chained_binary_expressions = chop_if_long
> resharper_wrap_object_and_collection_initializer_style = chop_always
> ```
> 
> I generated it with Rider's "Auto-Detect Formatting Rules" functionality, then tweaked it a bit further.

I totally agree with this, and at the moment I'll revert the suggested file.
When I removed this `.editorconfig` file I somehow in the middle of this PR I adopted Riders IDE which will automatically format code if (BIG if) developer/contributor is working within this IDE...but again, we didn't cover the rest of the world 🗺️ 
Hope it will fit now, but as this is first version for this huge change, I would be happy to fix/adopt any new PR/Issues regarding formating and its rules 💪 @jgiannuzzi

ljubon:
as agree, added `.editorconfig` to projects folder, in order to match format for future contributors 🙏

GPSnoopy:
Latest version of the PR, code formatting wise, looks quite good.

Duplicated copies of `.editorconfig` is rather sad. I wonder how we could solve this. Maybe have a copy script automatically run when the build scripts are called and also on CI?

There are a couple of massive PRs in the pipe I'd like to merge first before merging this one.

ljubon:
> Latest version of the PR, code formatting wise, looks quite good.
> 
> Duplicated copies of `.editorconfig` is rather sad. I wonder how we could solve this. Maybe have a copy script automatically run when the build scripts are called and also on CI?
> 
> There are a couple of massive PRs in the pipe I'd like to merge first before merging this one. So, let's aim to merge this once version 4.0.0-beta2 has been released.

Hey @GPSnoopy, thank you for the feedback!
Regarding the script and `.editorconfig` it's more practical, but since we are not running any formating over that file I would keep it as it is. The goal of `.editconfig` file is to `format` code locally on developer's machine before he pushes the code.
Once code is pushed, following the same rules we will check with `jb dotnet format` if the format is satisfying he should get build pass status.

GPSnoopy:
@ljubon I've enabled GitHub Actions for this PR. It should help you with the `CI/check-format` stage.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,27 @@ jobs:
 
   # Note that vcpkg dependencies takes the majority of the build time.
   # We cache them using GitHub Actions cache and export, making the scripts below a bit more complex.
+  check-format:
```
GPSnoopy, line 17: We really ought to move this to another file in a separate PR.

GPSnoopy (approved):

PR #201, opened 2021-06-15
Try ReSharper dotnet tool jb code formatting.

GPSnoopy:
Apply with `jb cleanupcode --profile="Built-in: Reformat Code" --settings="Formatting.DotSettings"  "csharp" "csharp.test" "csharp.benchmark"`


csharp.test/TestEncryption.cs
```diff
@@ -136,7 +136,7 @@ public static void TestEncryptJustOneColumn()
             {
                 using var fileReader = new ParquetFileReader(input);
                 using var groupReader = fileReader.RowGroup(0);
-                    
+
```
ljubon, line 139: I think we have different behavior when we run same `.DotSettings` file on my machine and yours...
I run `jb cleanup` with your example of `.DotSettings` file (given [here](https://github.com/G-Research/ParquetSharp/pull/196#issuecomment-861434901))
FYI I pushed new changes to my original https://github.com/G-Research/ParquetSharp/pull/196 with your file 

If we just compare one file from your run and mine...e.g `csharp.test/TestEncryption.cs` you'll see that my first change is at [line 20](https://github.com/G-Research/ParquetSharp/pull/196/files#diff-7270b6186d7960ee4b0c8cd04d9cda95b5cc5fe358ef8cb639b93a89affcc030) and from your run it's here at line 139 line 🤔 

Reason for this could be that you have your local (per user) .DotSettings which add additional rules for formating the code.... it's a wild guess but couldn't think another one

I'm running command on Windows10, Command prompt as terminal, but same result I have if I run it under WSL 2 (Ubuntu 20.04)
```
C:\Users\LJUBOOPS\work\G-Research\ParquetSharp>jb cleanupcode "csharp" "csharp.test" "csharp.benchmark" --profile="Built-in: Reformat Code" --settings="ParquetSharp.DotSettings" --verbosity=WARN
JetBrains Cleanup Code 2020.3
Running on AMD 64 in 64-bit mode, .NET runtime 3.1.15 under Microsoft Windows NT 10.0.19042.0

C:\Users\LJUBOOPS\work\G-Research\ParquetSharp>dotnet --version
5.0.300
```

jgiannuzzi:
Could you please add your `Formatting.DotSettings` to this branch?

csharp.test/TestEncryption.cs
GPSnoopy: It seems I have a more recent version of `jb`:

```
D:\Development\Work\ParquetSharp>jb cleanupcode --help
JetBrains Cleanup Code 2021.1.3
```

GPSnoopy:
> Could you please add your `Formatting.DotSettings` to this branch?

I thought I did. Not sure how I managed to commit everything but that one file. Fixed now.

csharp.test/TestEncryption.cs
ljubon: my bad, i run `jb cleanupcode` command, but actually I installed `jb` as standalone and as `dotnet tool restore` so when I run with `dotnet jb cleanupcode` i have same version, and still same reformating

ljubon: not it's matching :) 
this `.DotSettings` is different then one you shared in https://github.com/G-Research/ParquetSharp/pull/196 , could be format of the file 
anyway, you can take a look now at https://github.com/G-Research/ParquetSharp/pull/196 since changes should match yours 🤞

PR #202, opened 2021-06-17
Align release versions with Arrow's (4.0.0.0-beta1)

GPSnoopy:


GPSnoopy:
Unfortunately this doesn't look like it would be working :-\
https://docs.microsoft.com/en-us/nuget/concepts/package-versioning

4.0.0.0 would be read as 4.0.0

csharp/ParquetSharp.csproj
```diff
@@ -13,7 +13,7 @@
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <NoWarn>1591;</NoWarn>
-    <Version>2.5.0-beta1</Version>
+    <Version>4.0.0.0-beta1</Version>
```
jgiannuzzi, line 16: You should make it `4.0.0-beta1`.
If we make another release of ParquetSharp that is still using Arrow 4.0.0, it would become `4.0.0.1-beta1` and should work fine.

See [releases of Consul.NET](https://www.nuget.org/packages/Consul/) for example.

GPSnoopy: I'm confused. This is not what the doc above says

jgiannuzzi: See https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#normalized-version-numbers

> A zero in the fourth part of the version number will be omitted

So our first release that uses Arrow 4.0.0 should be `4.0.0`, and any other release that will use it should be `4.0.0.x` (where x > 1).

jgiannuzzi (approved):

PR #203, opened 2021-06-21
Make Date, DateTimeNanos IComparable

m00ngoose:


GPSnoopy:
Out of curiosity, what do you need this for?

IMHO typeless comparison is a bit of a code smell, as you rarely have heterogeneous collections of unknown `object`s. So my default stance would not to do this.

m00ngoose:
> Out of curiosity, what do you need this for?
> 
> IMHO typeless comparison is a bit of a code smell, as you rarely have heterogeneous collections of unknown `object`s. So my default stance would be not to do this.

I don't believe it to be a code smell. I want it because eg. a (int, Date) tuple compared with StructuralComparisons.StructuralComparer will call Comparer.Default.Compare which calls (non-generic) IComparable.Compare.

> Out of curiosity, what do you need this for?
> 
> IMHO typeless comparison is a bit of a code smell, as you rarely have heterogeneous collections of unknown `object`s. So my default stance would be not to do this.

It's needed if you ever want to compare eg. `Date[]` or `(int, Date)`s (StructuralComparisons.StructuralComparer calls Comparer.Default calls (untyped) IComparable.Compare).

GPSnoopy:
Off-line conversation with @m00ngoose has convinced me this is needed due to the row-oriented API which works with tuples. Not ideal, but better than nothing I guess.

GPSnoopy:
Actually, digging up further I still don't understand. `ValueTuple` uses `Comparer<T>.Default` which IIUC will use the typed `IComparable<T>` interface, so this change should not be needed when sorting `Date[]` or `(int, Date)[]`.

GPSnoopy:
Further clarification, .NET classes like `StructuralComparisons` provide no type-safe alternative that I'm aware of. In that context, this PR makes sense.

GPSnoopy (approved):

Issue #205, opened 2021-06-24
Column names not correctly translating UTF-8


CurtHagenlocher:
Create a test file from Python with

```
import pyarrow as pa
import pyarrow.parquet as pq
schema = pa.schema([
    ('Kein', pa.int32()),
    ('Ärger', pa.string())
])
column1 = pa.array([1, 2, 3], type = pa.int32())
column2 = pa.array(['ä', 'ö', 'ü'], type = pa.string())
batch = pa.RecordBatch.from_arrays([column1, column2], names = schema)
table = pa.Table.from_batches([batch])
pq.write_table(table, 'c:/testdata/test.parquet')

```

Then consume it with ParquetSharp

```
using (var reader = new ParquetFileReader("C:/testdata/test.parquet"))
{
    foreach (var node in reader.FileMetaData.Schema.GroupNode.Fields)
    {
        System.Console.WriteLine(node.Name);
    }
}
```

and the second column name is returned as "A,rger" instead of "Ärger". This is because Marshal.PtrToStringAnsi is being used, whereas I believe the column names are actually encoded as UTF-8.

The test file is also attached.

[Test.zip](https://github.com/G-Research/ParquetSharp/files/6711232/Test.zip)

GPSnoopy:
If that's the case, then quite a few API points need to be tested and modified: 
- node names
- column creation
- column names
- node paths

adamreeve:
I'm closing this as it was fixed with #206.

PR #206, opened 2021-06-28
Add proper UTF8 Support

GPSnoopy:
Fixes Issue #205.

Doing this PR requires `[MarshalAs(UnmanagedType.LPUTF8Str)]` which is only supported in net47 or higher. This means I've taken the decision to retire support for .NET Framework 4.6.1 and .NET Core 2.0 or lower. Otherwise, the code would be significantly more complicated and require explicit conversion of all strings to UTF8 before calling into the native methods. Even so, I had to backport `PtrToStringUTF8` from .NET Standard 2.1 (can't believe marshalling UTF8 wasn't properly supported until .NET Core 3.0).

The rest should be less controversial.

One area I'm not sure is `key_id` vs `key_metadata`. I think one of them is meant to be a byte array while the other should be a UTF8 string. But the documentation is lacking. This PR does not address this ambiguity (left out of scope on purpose).

### DIFF
--- a/csharp/StringUtil.cs
+++ b/csharp/StringUtil.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace ParquetSharp
+{
+    internal static class StringUtil
+    {
+        public static unsafe IntPtr ToCStringUtf8(string str, ByteBuffer byteBuffer)
+        {
+            var ascii = System.Text.Encoding.UTF8;
+            var byteCount = ascii.GetByteCount(str);
+            var byteArray = byteBuffer.Allocate(byteCount + 1);
+
+            fixed (char* chars = str)
+            {
+                ascii.GetBytes(chars, str.Length, (byte*)byteArray.Pointer, byteCount);
+            }
+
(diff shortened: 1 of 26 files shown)


csharp/StringUtil.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace ParquetSharp
+{
+    internal static class StringUtil
+    {
+        public static unsafe IntPtr ToCStringUtf8(string str, ByteBuffer byteBuffer)
+        {
+            var ascii = System.Text.Encoding.UTF8;
```
adamreeve, line 10: ascii?

GPSnoopy: Doh! 🤦

GPSnoopy: Fixed

adamreeve (approved):

jgiannuzzi (dismissed):
Thanks for the review @adamreeve! 🚀
And thanks for the PR @GPSnoopy!

jgiannuzzi (approved):

PR #208, opened 2021-06-29
Try to fix random segfault under Ubuntu 18.04 CI

GPSnoopy:
https://github.com/dotnet/sdk/issues/11639

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,8 @@ on:
   pull_request:
 
 env:
-  DOTNET_NOLOGO: true
+  DOTNET_CLI_TELEMETRY_OPTOUT: true
+  DOTNET_NOLOGO: true  
   dotnet_3_version: '3.1.407'
   dotnet_5_version: '5.0.201'
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
 
 env:
   DOTNET_CLI_TELEMETRY_OPTOUT: true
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -5,7 +5,8 @@ on:
   pull_request:
 
 env:
-  DOTNET_NOLOGO: true
+  DOTNET_CLI_TELEMETRY_OPTOUT: true
+  DOTNET_NOLOGO: true
```
jgiannuzzi, line 9: Can you fix the white space please?

GPSnoopy: Done. I blame GitHub web editor

GPSnoopy:
> I'm very happy to try out disabling the telemetry, but how come this would fix those segfaults we randomly get? 😳

Googling around I found links like https://github.com/dotnet/core/issues/973. The consensus seems to be that having the wrong or conflicting SSL libraries installed messes up with dotnet install.

So far I've observed that about 30-50% of our CI builds fail due to a segfault when calling `dotnet remove`. They always fail in the Ubuntu 18.04 images (but not 20.04). I suspect it's going to be something silly like this. May or may not solve the problem, but I have to try something.

I do miss having the ability to interactively log onto a GitHub action job/image.Would be super useful when things do go wrong.

GPSnoopy:
@jgiannuzzi Maybe something to be tried at some point? I can imagine your team wanting this: https://github.community/t/workspace-interactive-shell-or-workspace-browser/16498/4

PR #214, opened 2021-07-30
Update to Arrow-5.0.0

GPSnoopy:


cpp/ParquetFileWriter.cpp
```diff
@@ -90,8 +90,8 @@ extern "C"
 	{
 		TRYCATCH
 		(
-			const auto m = writer->key_value_metadata();
-			*key_value_metadata = m ? new std::shared_ptr<const KeyValueMetadata>(m) : nullptr;
+			const auto& m = writer->key_value_metadata();
+			*key_value_metadata = m ? new std::shared_ptr(m) : nullptr;
```
jgiannuzzi, line 94: I don't fully grasp what changing `const auto` to `const auto&` does, and why it's not done on line 102 below. Could you please explain? 😅

cpp/Enums.cpp
```diff
@@ -71,6 +72,7 @@ namespace
 		static_assert(Repetition::REQUIRED == 0);
 		static_assert(Repetition::OPTIONAL == 1);
 		static_assert(Repetition::REPEATED == 2);
+		static_assert(Repetition::UNDEFINED == 3);
```
jgiannuzzi, line 75: Whilst we're at it, we should add `Encoding::UNKNOWN` with value `999`.
See https://github.com/apache/arrow/blob/apache-arrow-5.0.0/cpp/src/parquet/types.h#L478

cpp/ParquetFileWriter.cpp
GPSnoopy: Probably one that was highlighted by ReSharper C++, as it's quite subtle.

`const auto&` means "a constant reference", whereas `const auto` implies a constant copy. For some reason, `writer->key_value_metadata()` returns a reference to a smart pointer, and `writer->metadata()` returns a smart pointer copy. 

Thus slight inconsistency within Arrow here (probably doesn't matter), but at least now our code correctly reflects this difference.

cpp/Enums.cpp
GPSnoopy: I've left it out since we currently do not expose `Encoding.Unknown` in C#. Thus no risk of ABI breaks here.

Also, a bit of a WTF here. What's the difference from Arrow's point of view between `Encoding::UNDEFINED` and `Encoding::UNKNOWN` ?

jgiannuzzi: Thanks, I get your point about ABI breaks. Why do we start exposing `Repetition.Undefined` in C# now though?

jgiannuzzi (approved):

PR #215, opened 2021-07-30
Promote 4.0.0-beta2 to stable release

GPSnoopy:


jgiannuzzi (commented):
I think we should use either `VCPKG_DISABLE_METRICS`, or `-disableMetrics`, but not both.
Setting the environment variable might be the most sensible choice, as it disables telemetry in the CI, but not when building locally (where users can choose to opt out by setting the env var themselves).

Also one could argue these changes should be part of a separate PR, but I know you came up with them whilst trying to solve the CI issues that were introduced by a bad runner virtual environment update GitHub made.

jgiannuzzi (approved):

PR #218, opened 2021-08-19
Add isSupportedType method back in LogicalTypeFactory

### LINKED ISSUE #217: Add IsSupported() method back, but on LogicalTypeFactory
The following code is quite useful to know whether a certain C# type is going to be accepted by ParquetSharp or needs user-side serialisation

```c#
        /// <summary>
        /// Query whether the given C# type is supported and a schema node can potentially be created.
        /// </summary>
        public static bool IsSupported(Type type)
        {
            if (type == null) throw new ArgumentNullException(nameof(type));

            while (true)
            {
                if (Primitives.ContainsKey(type))
                {
                    return true;
                }

                if (type.IsArray)
                {
                    type = type.GetElementType();
                    continue;
                }

                return false;
            }
        }
```

Could we add it back to `LogicalTypeFactory`?

damelLP:
closes #217

### DIFF
--- a/csharp.test/TestLogicalTypeFactory.cs
+++ b/csharp.test/TestLogicalTypeFactory.cs
@@ -173,15 +173,6 @@ namespace ParquetSharp.Test
             TestWriteNoColumnNorWriterOverride(ArrayValues, ArrayCustomValues);
         }
 
-        [Test]
-        public static void TestUnsupportedType()
-        {
-            Assert.False(LogicalTypeFactory.IsSupported(typeof(TestColumn)));
-
-            var exception = Assert.Throws<ArgumentException>(() => new Column<object>("unsupported").CreateSchemaNode());
-            Assert.AreEqual("unsupported logical type System.Object", exception?.Message);
-        }
-
         // Reader tests.
 
         private static void TestRead<TCustom, TValue>(TCustom[] expected, TValue[] written)
--- a/csharp.test/TestLogicalTypeFactory.cs
+++ b/csharp.test/TestLogicalTypeFactory.cs
@@ -173,6 +173,15 @@ namespace ParquetSharp.Test
             TestWriteNoColumnNorWriterOverride(ArrayValues, ArrayCustomValues);
         }
 
+        [Test]
+        public static void TestUnsupportedType()
+        {
+            Assert.False(LogicalTypeFactory.IsSupported(typeof(TestColumn)));
+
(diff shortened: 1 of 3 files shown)


csharp/LogicalTypeFactory.cs
```diff
@@ -34,6 +34,76 @@ public virtual (Type physicalType, Type logicalType) GetSystemTypes(ColumnDescri
...
+                {typeof(TimeSpanNanos), (Repetition.Required, LogicalType.Time(isAdjustedToUtc: true, timeUnit: TimeUnit.Nanos), PhysicalType.Int64)},
+                {typeof(TimeSpanNanos?), (Repetition.Optional, LogicalType.Time(isAdjustedToUtc: true, timeUnit: TimeUnit.Nanos), PhysicalType.Int64)},
+                {typeof(string), (Repetition.Optional, LogicalType.String(), PhysicalType.ByteArray)},
+                {typeof(byte[]), (Repetition.Optional, LogicalType.None(), PhysicalType.ByteArray)}
+            };
+
+        /// <summary>
+        /// Query whether the given C# type is supported and a schema node can potentially be created.
+        /// </summary>
+        public static bool IsSupported(Type type)
```
GPSnoopy, line 85: Suggest a non-static `virtual` method instead (then user-side it can be called on `LogicalTypeFactory.Default` where you need it, or on the `LogicalTypeFactory` instance of the `ParquetFileReader` if the latter is given).

That way you don't need the static dictionary above either.

csharp.test/TestLogicalTypeFactory.cs
```diff
@@ -173,6 +173,15 @@ public static void TestWriteNoColumnNorWriterOverride_Array()
             TestWriteNoColumnNorWriterOverride(ArrayValues, ArrayCustomValues);
         }
 
+        [Test]
```
GPSnoopy, line 176: Suggest you remove this test. One already exist in `TestColumn.cs`.

In fact, https://github.com/G-Research/ParquetSharp/pull/185/files removes the two calls to `IsSupported`. We can just add them back in both places.

csharp/LogicalTypeFactory.cs
damelLP: fair enough do you envision the dictionary being passed as a constructor arg instead?

GPSnoopy: `LogicalTypeFactory` already has a constructor taking a dictionary in. Hence why you don't need to worry about it.

GPSnoopy (approved):

PR #223, opened 2021-11-01
Bump version of dotnet 5 SDK used in CI

adamreeve:
This should fix "The repository primary signature validity period has expired" errors on macOS.

See https://github.com/dotnet/core/issues/6595


jgiannuzzi (dismissed):
Good spot, thanks @adamreeve!

I suggest we take advantage of this change to update to the latest SDKs: 3.1.414 and 5.0.402

jgiannuzzi (approved):

PR #224, opened 2021-11-05
Automatic NuGet and release publication

jgiannuzzi:
This PR enables the automatic creation of a draft GitHub release with the NuGet Package when a tag is created, as well as uploading the package to nuget.org.

#### Pre-requisites for merging
- [x] creation of `NUGET_API_KEY` secret with upload rights to nuget.org

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -249,8 +249,6 @@ jobs:
         dotnet add csharp.test package ParquetSharp -v ${{ steps.get-version.outputs.version }}
     - name: Build & Run .NET unit tests
       run: dotnet test csharp.test --configuration=Release --framework ${{ steps.dotnet.outputs.framework }}
-    outputs:
-      version: ${{ steps.get-version.outputs.version }}
 
   # Create a GitHub release and publish the NuGet packages to nuget.org when a tag is pushed.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -249,3 +249,45 @@ jobs:
         dotnet add csharp.test package ParquetSharp -v ${{ steps.get-version.outputs.version }}
     - name: Build & Run .NET unit tests
       run: dotnet test csharp.test --configuration=Release --framework ${{ steps.dotnet.outputs.framework }}
+    outputs:
+      version: ${{ steps.get-version.outputs.version }}
```
adamreeve, line 253: Is this used? I was expecting to see this version used in the publish-release action but it looks like that computes the version itself

adamreeve (approved):

.github/workflows/ci.yml
jgiannuzzi: No, it's not used 😅
I was planning to use it to avoid recomputing it below, but decided against it because I am planning a future refactor where `publish-release` won't have a direct dependency on `test-nuget`.

Thanks a lot for spotting it!

adamreeve (approved):

PR #230, opened 2021-11-12
Update to Arrow-6.0.0

### LINKED ISSUE #222: Arrow 6.0 has been released
* Parquet version enums have changed.
* Null statistics might have been fixed in more cases (check unit tests)
* Decimal statistics might have been fixed in more cases (#68 and unit tests)

jgiannuzzi:
Closes #222.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,8 @@ if (MSVC)
 
 	add_compile_options("/MP") # Compile files in parallel.
 	add_compile_options("/WX") # Threat warnings as errors.
+	# See https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
+	add_compile_options("/Zc:__cplusplus") # Enable standard __cplusplus macro.
 
 endif ()
 
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ In the 5.0.X versions, reading nested structures was introduced. However, nestin
 
 Building ParquetSharp for Windows requires the following dependencies:
 - Visual Studio 2019 (16.4 or higher)
-- Apache Arrow (5.0.0)
+- Apache Arrow (6.0.0)
 
 For building Arrow (including Parquet) and its dependencies, we recommend using Microsoft's [vcpkg](https://github.com/Microsoft/vcpkg). Note that the Windows build needs to be done in a Visual Studio x64 Native Tools Command Prompt for the build script to succeed.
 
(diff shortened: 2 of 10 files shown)


CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ if (MSVC)
 
 	add_compile_options("/MP") # Compile files in parallel.
 	add_compile_options("/WX") # Threat warnings as errors.
+	add_compile_options("/Zc:__cplusplus") # Enable standard __cplusplus macro.
```
GPSnoopy, line 36: Why is this part needed ?

jgiannuzzi: Without this, the build on Windows fails because the `__cplusplus` macro is not defined by default by MSVC: https://github.com/G-Research/ParquetSharp/runs/4137772529?check_suite_focus=true#step:18:100

GPSnoopy: Highly suspicious. This macro should always be defined for C++ compilation:
https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170&viewFallbackFrom=vs-2017

I think the problem comes from arrow in arrow/cpp/src/arrow/util/launder.h:
`#if __cplusplus >= 201703L`

I don't think that the actual value of __cplusplus can be relied on in MSVC. For example, even though I'm compiling for C++20, I get `__cplusplus = 199711` in Visual Studio 2022.

Looks like you came to the same conclusion:
/Zc:__cplusplus : Enable the __cplusplus macro to report the supported standard (off by default).

I suggest you raise this with Ian.

jgiannuzzi: Yes, that is the exact reason I had to define `/Zc:__cplusplus`. It's quite unfortunate, but Visual Studio will define `__cplusplus` wrongly by default (See https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/).

I did mention it to Ian [here](https://github.com/microsoft/vcpkg/pull/21113#issuecomment-967006107), though I'm not sure this is a problem with the vcpkg port of Arrow, but rather a quirk of MSVC that we were lucky enough to not encounter until Arrow 6.0.0.

GPSnoopy: Worth adding this URL as a comment above this line in the CMake file then

jgiannuzzi: Great point, thanks!

jgiannuzzi:
Closed in favour of #247.

PR #231, opened 2021-11-19
Enable support for arm64

### LINKED ISSUE #228: Support for macOS - Arm64
Now with .NET 6 out, would it be possible to add native support for macOS Arm64 / M1 processors?

jgiannuzzi:
Fixes #228 

This PR enables support for the arm64 architecture (aka aarch64 on Linux, or Apple Silicon on macOS).
This is possible thanks to https://github.com/microsoft/vcpkg/pull/21467 that has been merged in vcpkg. Support for arm64 on Windows is not yet enabled, as it requires further changes to vcpkg.

The CI pipeline is updated to build the native libs for arm64 on Linux and macOS, and to test the arm64 Linux build through emulation. The arm64 macOS build has been tested locally on a M1 Mac, but cannot yet be tested through the CI (waiting for macstadium.com to enable our Open Source account).

Caveats: the NuGet package size goes from 18MB to 29MB, and running the tests on emulated arm64 takes almost 20 minutes.


.github/workflows/ci.yml
```diff
@@ -47,13 +48,17 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04, macos-latest, windows-latest]
```
pavlovic-ivan, line 50: Can `windows-latest` be removed from the os list, since there is an `exclude` of the `windows-latest` below?

adamreeve (approved):
:+1: Looks good to me. 20 minutes seems like a long time to run the tests. It could be worth checking if there's any easy way to reduce the runtime (eg. can we reduce test data sizes for the longest running tests?). That would probably be a separate PR though

mfkl:
> 20 minutes seems like a long time to run the tests.

Indeed, that's 20 times longer than the other runs. These 20 minutes include pulling the docker image though (not sure how fast that is on GH).

jgiannuzzi:
> These 20 minutes include pulling the docker image though (not sure how fast that is on GH).

Pulling the docker image seems pretty fast — it's really the emulation of the restore+build+test that's killing the performance (unsurprisingly 😞)
@pavlovic-ivan is currently working on a framework to spin up self-hosted runners dynamically — when ready, this could be used to spin up a real arm64 linux runner and run the tests at native speed.

PR #238, opened 2021-12-01
Statistics test updates for Arrow 6.0.0

adamreeve:
* Re-enables checking of null counts, which were previously broken for array valued columns
* Re-enables checking min/max for decimals, which has been fixed since Arrow 4.0


jgiannuzzi (approved):

jgiannuzzi:
Thanks a lot @adamreeve!

Issue #242, opened 2022-01-06
ParquetSharp cannot read arrays for large files.


adeboyed:
Reading string arrays fails for files which have n rows, where n is > 4096 rows and not a multiple of 4096.
This also occurs for uint arrays, I haven't however tested every Parquet supported type.

### Steps to reproduce
1. Create a Parquet file with a column of string arrays, e.g. using pandas:
```
Python:

import pandas as pd

data = {'data': list(map(lambda x: [f"Row no: {x}"], range(4100))) }
pd.DataFrame(data).to_parquet(r"C:\test.parquet")
```
2. Read Parquet file using ParquetSharp
``` 
C#: 

using var parquetReader = new ParquetFileReader("C:\\test.parquet");
using var rowGroupReader = parquetReader.RowGroup(0);
using var dataColumn = rowGroupReader.Column(0).LogicalReader<string[]>().GetEnumerator();

Console.WriteLine($"No of rows : {parquetReader.FileMetaData.NumRows}");
for (var i = 0; i < rowGroupReader.MetaData.NumRows; i++)
{
   var moveNext = dataColumn.MoveNext();
   Console.WriteLine($"moveNext: {moveNext} data: {dataColumn.Current[0]}");
}
```

**Output:**
```
No of rows : 4100

MoveNext: 'True' | Value: 'Row no: 1'
MoveNext: 'True' | Value: 'Row no: 2'
MoveNext: 'True' | Value: 'Row no: 3'
...
MoveNext: 'True' | Value: 'Row no: 4091'
MoveNext: 'True' | Value: 'Row no: 4092'
MoveNext: 'True' | Value: 'Row no: 4093'
MoveNext: 'True' | Value: 'Row no: 4094'
MoveNext: 'True' | Value: 'Row no: 4095'
MoveNext: 'False' | Value: 'Row no: 4095'
MoveNext: 'False' | Value: 'Row no: 4095'
MoveNext: 'False' | Value: 'Row no: 4095'
MoveNext: 'False' | Value: 'Row no: 4095'
```
As this file has 4100 rows, the value of the single column should keep incrementing until 4100, however after 4096 rows, the enumerator stops returning true for MoveNext(), and ParquetSharp does not return the remaining values.

This is likely related to the default buffer length of 4096, but I haven't properly looked into it it: https://github.com/G-Research/ParquetSharp/blob/39acb40e99bb8e8af5f01d678acefee43d45fbfb/csharp/ColumnReader.cs#L75

[...]

Issue #229, opened 2021-11-10
ColumnReader.Skip can give access violation for Nullable<bool> field


TobyShaw:
Code repro: 

```fsharp
open ParquetSharp
open ParquetSharp.IO

let resizableBuffer = new ResizableBuffer()

let column = Column<bool Nullable> "Value"
do
    // Set-up: Make a file with one row-group, one column, the column is filled with 100,000 nullable bools.
    // Note about F# syntax: once we exit the do-block, all writers will have been disposed in the correct order.
    use outputStream = new BufferOutputStream(resizableBuffer)
    use fileWriter = new ParquetFileWriter(outputStream, [|column :> Column|])
    use rowGroup = fileWriter.AppendRowGroup()
    use logicalWriter = rowGroup.NextColumn().LogicalWriter<bool Nullable>()
    logicalWriter.WriteBatch [|for i in 1..100_000 -> Nullable (i % 2 = 0)|]

// Get a column reader for this Nullable<bool> column
use fr = new ParquetFileReader(new BufferReader(resizableBuffer))
use rgr = fr.RowGroup 0
let columnReader = rgr.Column(0)

// When we skip to halfway through the column, we get an access violation very consistently.
columnReader.Skip(50_000L) |> printfn "We crash before reaching this: %A"
```
I can reproduce this error with a file on disk, as well as an in-memory buffer (it just happens to be easier to give a repro using the in-memory buffer)

This was observed on versions 2.2.0 and 5.0.0-beta1 of ParquetSharp

jgiannuzzi:
Thanks for your bug report, @TobyShaw!

We will look into this.

adamreeve:
Hi @TobyShaw, this is a bug in the underlying Apache Arrow library (see https://issues.apache.org/jira/browse/ARROW-11518). This has been fixed in Arrow 6.0.0, so should be fixed by #230.

jgiannuzzi:
Fixed in [ParquetSharp 6.0.1-beta1](https://github.com/G-Research/ParquetSharp/releases/tag/6.0.1-beta1)

Issue #68, opened 2019-08-06
Decimal statistics seem to be wrong


philjdf:
https://issues.apache.org/jira/browse/ARROW-6149

jgiannuzzi:
Fixed in [ParquetSharp 6.0.1-beta1](https://github.com/G-Research/ParquetSharp/releases/tag/6.0.1-beta1)

PR #244, opened 2022-01-11
Fix GetEnumerator

philjdf:
Would close https://github.com/G-Research/ParquetSharp/issues/242

### DIFF
--- a/csharp.test/TestLogicalTypeRoundtrip.cs
+++ b/csharp.test/TestLogicalTypeRoundtrip.cs
@@ -288,14 +288,19 @@ namespace ParquetSharp.Test
             fileReader.Close();
         }
 
+        /// <summary>
+        /// This checks that LogicalColumnReader's GetEnumerator() works correctly
+        /// when the column is longer than the buffer length but not an exact multiple
+        /// (see https://github.com/G-Research/ParquetSharp/issues/242).
+        /// </summary>
         [Test]
         public static void TestLargeArraysEnumerator()
         {
-            CheckEnumerator(Enumerable.Range(0, 4100).ToArray());
-            CheckEnumerator(Enumerable.Range(0, 4100).Select(i => new[] {$"row {i}"}).ToArray());
+            CheckEnumerator(4096, Enumerable.Range(0, 4100).ToArray());
+            CheckEnumerator(4096, Enumerable.Range(0, 4100).Select(i => new[] {$"row {i}"}).ToArray());
         }
 
-        private static void CheckEnumerator<T>(T[] values)
+        private static void CheckEnumerator<T>(int bufferLength, T[] values)
         {
             using var buffer = new ResizableBuffer();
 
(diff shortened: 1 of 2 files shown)


csharp.test/TestLogicalTypeRoundtrip.cs
```diff
@@ -288,6 +288,49 @@ public static void TestNestedStructArray([Values(Repetition.Required, Repetition
             fileReader.Close();
         }
 
+        [Test]
+        public static void TestLargeArraysEnumerator()
+        {
+            CheckEnumerator(Enumerable.Range(0, 4100).ToArray());
+            CheckEnumerator(Enumerable.Range(0, 4100).Select(i => new[] {$"row {i}"}).ToArray());
```
adeboyed, line 295: This test works on the assumption that the default BufferLength=4096. 
Might be worth explicitly setting that.

philjdf: Good idea, specified the buffer length to use and added a comment.

adamreeve (approved):

jgiannuzzi:
Given the ABI change in this PR, and given that 5.0.0-beta3 has now been out and stable for a while, I would like to release 5.0.0 first, and then release 6.0.0-beta1 immediately after with the upgrade to Arrow 6.0.0 _and_ this PR.

@philjdf @adamreeve @adeboyed what are your thoughts on that?

adeboyed:
Could this PR be included in the 5.0.0 release since upgrading to Arrow 6 may potentially lead to extended testing to ensure that the new features are supported and any API changes are made.

jgiannuzzi:
> Upgrading to Arrow 6 may potentially lead to extended testing to ensure that the new features are supported and any API changes are made.

If you are talking about ParquetSharp itself, then there are no API changes stemming from the upgrade to Arrow 6, and no new features. All the work has already been done in #230 and is just waiting for 5.0.0 stable to be released.
Let me know if you meant something else!

> Could this PR be included in the 5.0.0 release?

If you mean 5.0.0 stable, then I would be quite reluctant in having an ABI change included without releasing another beta first. We _could_ however include it in another 5.0.0 beta. This would push the work on the 6.0.0 series further however, which I'm not really keen on, especially as #68 and #229 would be fixed by it.

jgiannuzzi:
@adeboyed could you please let me know today if my comment above makes sense to you — if yes, I'll go ahead and release 5.0.0 and 6.0.0-beta1 (@philjdf and @adamreeve already agreed offline with this approach)

adeboyed:
Hi Jonathan, this approach makes sense to me.
As an aside, we should change the comment here: https://github.com/G-Research/ParquetSharp/blob/8ce59329580e01a0e98eb157a56df4a43cfefcfc/README.md?plain=1#L145 to reflect that this logic has made it into a release version.

jgiannuzzi:
Good catch @adeboyed! I made that change in PR #246 and released 5.0.0 stable.

jgiannuzzi:
The upgrade to Arrow 6.0.1 has also been merged (PR #247).

jgiannuzzi:
[ParquetSharp 6.0.1-beta1](https://github.com/G-Research/ParquetSharp/releases/tag/6.0.1-beta1) has been released with this fix.

PR #247, opened 2022-01-14
Update to Arrow 6.0.1

### LINKED ISSUE #68: Decimal statistics seem to be wrong

### LINKED ISSUE #222: Arrow 6.0 has been released

### LINKED ISSUE #229: ColumnReader.Skip can give access violation for Nullable<bool> field

jgiannuzzi:
Closes #222.
Closes #68.
Closes #229.
Supersedes #230.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,8 @@ if (MSVC)
 
 	add_compile_options("/MP") # Compile files in parallel.
 	add_compile_options("/WX") # Threat warnings as errors.
+	# See https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
+	add_compile_options("/Zc:__cplusplus") # Enable standard __cplusplus macro.
 
 endif ()
 
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ In the 5.0.X versions, reading nested structures was introduced. However, nestin
 
 Building ParquetSharp for Windows requires the following dependencies:
 - Visual Studio 2019 (16.4 or higher)
-- Apache Arrow (6.0.0)
+- Apache Arrow (6.0.1)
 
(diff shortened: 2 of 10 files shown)


marcin-krystianc (approved):
👍

PR #258, opened 2022-02-24
Support Windows Server 2022 CI builds

adamreeve:
GitHub has started rolling out Windows Server 2022 to GitHub actions jobs running on `windows-latest`: https://github.blog/changelog/2022-01-11-github-actions-jobs-running-on-windows-latest-are-now-running-on-windows-server-2022/

This image includes Visual Studio 2022 instead of 2019, so I've updated the build script to expect VS 2022, and explicitly set the Windows image to windows-2022 in the CI configuration so we're pinned to one version rather than randomly getting either 2019 or 2022 during the roll out.

Alternatively we could just stick with 2019, but I think it's best to keep up to date.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,10 +47,10 @@ jobs:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id
     strategy:
       matrix:
-        os: [ubuntu-18.04, macos-latest, windows-latest]
+        os: [ubuntu-18.04, macos-latest, windows-2022]
         arch: [x64, arm64]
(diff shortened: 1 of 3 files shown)


marcin-krystianc (approved):

jgiannuzzi (changes requested):
I wholeheartedly agree with keeping up to date, thanks @adamreeve!

We should also update the [Building section of the README](https://github.com/G-Research/ParquetSharp#building) to VS 2022, otherwise LGTM 👍

adamreeve:
Ah good catch thanks Jonathan, I missed that. I've updated the README now.

jgiannuzzi (approved):

PR #261, opened 2022-03-14
Read DateTime values with Kind set based on isAdjustedToUtc

### LINKED ISSUE #260: ParquetSharp should take into account "isAdjustedToUTC" for Timestamps
Quality of life improvement.

ParquetSharp by default returns DateTimes with [`DateTime.Kind`](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.kind?view=net-6.0) equal to `Unspecified`.
ParquetSharp can be a bit smarter and use `isAdjustedToUTC` field from the Parquet Format to specify the `DateTime.Kind` as `DateTime.Utc`

By default, ParquetSharp uses `new DateTime()` which returns an `DateTime.Kind.Unspecified` variant.

adamreeve:
Fixes #260 

This changes behaviour and also breaks the ABI for LogicalRead so should be released with a major version bump.


jgiannuzzi (approved):
LGTM 👍

Let's get 6.0.1 released before merging.

PR #262, opened 2022-03-21
Change DateTimeNanos.ToString format and allow overriding the format

adamreeve:
This PR changes the behaviour of `DateTimeNanos.ToString` so that by default it will use a custom format with nanosecond precision. It also adds the ability to override the formatting string and provide an `IFormatProvider`. 

I'm not sure whether this should be considered a breaking change and so only be released with a new major version. It isn't breaking from an API or ABI point of view, but changes behaviour for existing code using `DateTimeNanos.ToString`.


grlrrchapman (approved):
LGTM

jgiannuzzi:
Looks good to me 👍

I do think this could be seen as a (minor) breaking change, so I wonder whether we merge it and release as 6.0.1-beta2, or — given that we should upgrade to Arrow 7.0.0 — we release 6.0.1, upgrade Arrow, merge this, and release 7.0.0-beta1. Thoughts?

adamreeve:
Yeah I don't think this is particularly urgent (feel free to disagree @grlrrchapman) so can probably wait to go out with the 7.0.0 release.

grlrrchapman:
> Yeah I don't think this is particularly urgent (feel free to disagree @grlrrchapman) so can probably wait to go out with the 7.0.0 release.
Yeah this is nice to have, not urgent.

PR #264, opened 2022-03-28
Handle partial stream reads in ManagedRandomAccessFile

### LINKED ISSUE #263: Exception on partial read of stream: tried reading x bytes starting at position y from file but only got z
I am trying to read a parquet file from the cloud (Azure Data Lake). The stream is seekable, so I do it like suggested in #259 

I get a `ParquetException` though. It seems like sometimes only partial reads are performed. I tried the same operation in Parquet.NET and it worked.

Here is the exception:

```
Message
class parquet::ParquetException (message: ’Tried reading 11036647 bytes starting at position 2848913 from file but only got 4194304’)

Stack Trace
at ParquetSharp.ExceptionInfo.Check(IntPtr exceptionInfo)
at ParquetSharp.ExceptionInfo.Return[TArg0,TValue](IntPtr handle, TArg0 arg0, GetFuction`2 getter)
at ParquetSharp.ExceptionInfo.Return[TArg0,TValue](ParquetHandle handle, TArg0 arg0, GetFunction`2 getter)
at ParquetSharp.RowGroupReader.Column(Int32 i)
```

The part where you perform the read seems to forward the number of bytes read correctly: https://github.com/G-Research/ParquetSharp/blob/ca3d372febfb3ec3b1b3b737096164c446183e0d/csharp/IO/ManagedRandomAccessFile.cs#L47-L58

I guess the underlying C++ library does not support partial reads. It could be handled by `ManagedRandomAccessFile` by repeating reads until the buffer is full.

adamreeve:
Fixes #263 

I also noticed that the code within the `#if NETSTANDARD20` block was invalid and so not actually used. There needs to be an underscore in the version, and we're targeting .net standard 2.1 rather than 2.0 anyway (plus `Span<T>` was only added in 2.1). I've also switched to the `_OR_GREATER` version of the symbol for forwards compatibility if we upgrade the target framework later. These symbols are available since the 5.0 SDK (https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-a-target-framework).

`ManagedRandomAccessFile` and `ManagedOutputStream` should also now correctly handle reads and writes of more than the maximum byte array dimension, although I couldn't find a way to test this easily so possibly the underlying Arrow library won't request reads or writes this large anyway.


adamreeve:
The build failure looks like it needs a newer vcpkg version to resolve, as the version of libwinpthread that we're trying to download is no longer available and this should be fixed by https://github.com/microsoft/vcpkg/pull/22974.

The Arrow version is currently coupled to the vcpkg version though, which makes it tricky to fix this without bumping the Arrow version. We should probably be looking at migrating to using a vcpkg manifest so we can decouple the Arrow version from the vcpkg version: https://github.com/microsoft/vcpkg/blob/master/docs/users/manifests.md

mfkl:
Non-blocking, FYI remark.

> plus Span<T> was only added in 2.1

Indeed, but Span has been backported through the [System.Memory](https://www.nuget.org/packages/System.Memory/) package to older TFMs, so you can actually use it with netstandard1.1 and even net45. But the stream span overload is indeed missing (The System.IO package has not been updated in a long time). 

You could provide [this](`https://github.com/dotnet/runtime/blob/c4b9dabec8186a0d61f0cc3ea0b7efea579bf24e/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs#L720-L734`) as an extension method, but don't know if that is worth it here (for libvlcsharp, it was, I believe). Just FYI!

adamreeve:
Ah interesting, thanks @mfkl. Looks like we already have a dependency on System.Memory and use Span in quite a few other places, so yes it's only the Stream methods that are the reason for the `#if`, not the use of Span.

PR #265, opened 2022-03-28
Switch to using a vcpkg manifest to decouple Arrow version from vcpkg version

adamreeve:
The version of Arrow to build is currently coupled to the commit of vcpkg to check out, but this means we can't use a more up to date vcpkg with an older Arrow version, eg. to pull in a fix for a different dependency, see https://github.com/G-Research/ParquetSharp/pull/264#issuecomment-1080089710 for a case where this would be useful.

Vcpkg has introduced the concept of a [manifest file](https://github.com/microsoft/vcpkg/blob/master/docs/users/manifests.md), which allows specifying minimum version dependencies or overriding versions, alongside a baseline vcpkg version to use. It seems like migrating to using a manifest would make sense for ParquetSharp.

This has required a couple of other changes:
* Building with cmake will automatically build dependencies from vcpkg, so it doesn't make sense to have a separate `vcpkg install` step
* `vcpkg export` isn't yet compatible with a manifest, so I've switched the CI build to use caching based on vcpkg's binary source caching instead, which requires always downloading and configuring vcpkg even if we have a cache hit


vcpkg_version.txt
```diff
@@ -1 +1 @@
-https://github.com/microsoft/vcpkg.git 0fcee5a2ab497b851f6dbf55dc052bcff2b05072
+https://github.com/microsoft/vcpkg.git 9ff4659a075d5f4f30aaca9f7c4e7f059ecc1d6d
```
adamreeve, line 1: I'm not sure whether it makes sense to keep this specified separate to the `builtin-baseline` in `vcpkg.json`. I guess we could potentially want to use a new version of the vcpkg repo but keep an older baseline commit for versioning.

vcpkg.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg/master/scripts/vcpkg.schema.json",
+  "name": "parquetsharp",
+  "version": "5.0.0",
```
adamreeve, line 4: I don't think the version here really matters, as far as I can tell it's only important if we're publishing ParquetSharp itself as a vcpkg package. I guess we should either have a process to keep it up to date, or set it to something obviously wrong like `0.0.0`

marcin-krystianc (commented):
Looks reasonable, I've triggered few builds to see whether caching still works and it seems it works fine https://github.com/G-Research/ParquetSharp/actions/runs/2058041205

vcpkg.json
```diff
@@ -0,0 +1,15 @@
...
+  "builtin-baseline":"9ff4659a075d5f4f30aaca9f7c4e7f059ecc1d6d",
```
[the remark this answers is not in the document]
jgiannuzzi: if we switch to `vcpkg-configuration.json`, then we can get rid of `builtin-baseline`.

jgiannuzzi (commented):
This is awesome, Adam! 🚀

We should probably get rid of the `vcpkg_*` scripts and instead either assume `vcpkg` is installed in the `build_*` scripts (based on the `VCPKG_INSTALLATION_ROOT` env var), or install the latest version locally when not detected.

adamreeve:
Thanks for the feedback Jonathan and Marcin! I've removed the `vcpkg_*` scripts and moved to using `vcpkg-configuration.json`. I've also changed to using overlay triplets rather than modifying the triplet files in the vcpkg directory.

I'm checking for an existing vcpkg by whether `VCPKG_INSTALLATION_ROOT` or `VCPKG_ROOT` are defined. I could possibly also check for for vcpkg in `PATH`, eg. `VCPKG_INSTALLATION_ROOT=$(dirname $(which vcpkg))`, but gave up trying to figure out how to do that in cmd. Migrating `build_windows.bat` to PowerShell might be best if we want to do anything more complicated there...

jgiannuzzi:
@adamreeve I am _very_ much in favour of moving to PowerShell for the Windows build, even if we don't need to do anything complex!

jgiannuzzi:
@adamreeve I made a few trivial changes and pushed them to your branch — let me know if anything is unclear!

adamreeve:
Thanks @jgiannuzzi, nice catch noticing the overlay triplets weren't being written in the Ubuntu run, I should have picked up on that.

I've ported the Windows build script to PowerShell, although decided against trying to infer the location of vcpkg if there's a vcpkg command on PATH, as this might run into issues handling things like aliases, and it's possibly better to make users be explicit about whether to use an existing vcpkg by setting an environment variable. 

I was a bit confused at first about why the build that ran for f22861cfd0ab2450d72d4a3cf3c00ce495bfaa78 didn't use the cached files, but I'm guessing that the GitHub cache action doesn't update cached data after a cache hit. When the build for f61a832a63b03c71a420e8d3b4541ab516176a22 ran, it had a cache hit but was using the overlay triplets so had to recompile, and the updated binaries weren't re-cached.

jgiannuzzi:
Nice work porting to PowerShell! I agree about making it the location of vcpkg explicit for users.

About caching — you're right that the GitHub cache doesn't update cached data after an exact hit. It generally makes sense and our specific case of using overlay triplets without changing the cache key should not happen in real life. However, this got me thinking about when we want to update to a different baseline or version of Arrow, and I think we want to use [restore keys](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) as well. This would allow to have a partial match and download the 'closest' cache, then update it — thus speeding up the process when most dependencies haven't changed!

jgiannuzzi:
I've also merged the vcpkg triplet and root compute/resolve steps, then bumped the baseline again to validate our restore keys are working — it looks like it does and is giving a nice speedup. Unfortunately boost-modular-build-helper was changed in that commit, which meant rebuilding most packages 😅

jgiannuzzi (approved):
LGTM 👍

marcin-krystianc (approved):
Looks good 👍

PR #268, opened 2022-04-14
Upgrade to Arrow 7.0.0

jgiannuzzi:
This PR upgrades ParquetSharp to Arrow 7.0.0.

### Caveats

In order to get `ParquetSharpNative` to build on Windows, I had to narrow down all of our `parquet` imports.
Otherwise, we would get errors similar to this:
```
include/arrow/util/aligned_storage.h(37,1): error C2872: 'arrow': ambiguous symbol
include/parquet/types.h(33,11): message : could be 'arrow'
include/parquet/type_fwd.h(82,11): message : or       'parquet::arrow'
include/arrow/util/aligned_storage.h(36): message : while compiling class template member function 'T *arrow::internal::AlignedStorage<T>::get(void) noexcept'
          with
          [
              T=std::shared_ptr<arrow::DataType>
          ]
include/arrow/util/aligned_storage.h(50): message : see reference to function template instantiation 'T *arrow::internal::AlignedStorage<T>::get(void) noexcept' being compiled
          with
          [
              T=std::shared_ptr<arrow::DataType>
          ]
include/arrow/result.h(435): message : see reference to class template instantiation 'arrow::internal::AlignedStorage<T>' being
compiled
          with
          [
              T=std::shared_ptr<arrow::DataType>
          ]
include/arrow/type.h(1026): message : see reference to class template instantiation 'arrow::Result<std::shared_ptr<arrow::DataType>>' being compiled
```

Trying to compile something as simple as the following file on Windows reproduces the issue:
```cpp
#include <parquet/properties.h>

using namespace parquet;
```

This may be a bug in Arrow 7.0.0, and narrowing down the imports is the only workaround I found to be able to upgrade ParquetSharp. If anyone has a more elegant idea, please let me know!


adamreeve (approved):
Looks good to me. Not sure what would have changed to make the Windows build fail, but narrowing down the using statements doesn't seem like a bad thing.

marcin-krystianc:
> If anyone has a more elegant idea, please let me know!

I'm pretty rusty in c++ nowadays, so I might be wrong, but it seems to be a problem with how msvc compiler interprets templates. I would expect that adding `using` directive after `#include` directives shouldn't interfere with those headers included before, but apparently it is 🛸 

Apparently that behaviour is fixed in newer standard or msvc implementation of newer standard, so when the C++ 20 is used then the problem with name ambiguity doesn't occur anymore:

**In CMakeLists.txt:**
```
set (CMAKE_CXX_STANDARD 20)
```

Updating to C++ 20 standard is safe operation in mu opinion as we export only C-like API anyway.

jgiannuzzi:
Thanks @marcin-krystianc, that's a great find!

I've reverted the narrowing down and updated to C++ 20, which is much more elegant 😄

adamreeve (approved):

marcin-krystianc (approved):

PR #270, opened 2022-04-22
Fix vcpkg binary caching in CI

jgiannuzzi:
The `vcpkg` binary caching in our CI currently gets [broken](https://github.com/G-Research/ParquetSharp/runs/6121033684?check_suite_focus=true) when a cache entry gets [generated](https://github.com/G-Research/ParquetSharp/runs/6121033684?check_suite_focus=true) with a given version of `cmake` and then gets [used](https://github.com/G-Research/ParquetSharp/runs/6121033684?check_suite_focus=true#step:4:18) with a different version, e.g. when the runner virtual environment gets upgraded (`macOS-11` [20220410.1](https://github.com/actions/virtual-environments/blob/macOS-11/20220410.1/images/macos/macos-11-Readme.md#tools) vs [20220419.3](https://github.com/actions/virtual-environments/blob/macOS-11/20220419.3/images/macos/macos-11-Readme.md#tools)).
The broken cache entry then keeps on getting used for every subsequent run, resulting in a much slower CI feedback loop.

This PR aims at fixing this by adding the `cmake` version to the cache key.

Unrelated to this, the artifact names got mangled together during #265 (specifically my commit f78df21). This PR fixes that too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,7 +132,7 @@ jobs:
       if: steps.get-cached-vcpkg.outputs.cache-hit != 'true' && (success() || failure())
       uses: actions/upload-artifact@v2
       with:
-        name: ${{ steps.vcpkg-triplet.outputs.triplet }}-vcpkg-arrow-logs
+        name: ${{ steps.vcpkg-info.outputs.triplet }}-vcpkg-arrow-logs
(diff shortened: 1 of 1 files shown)


adamreeve (approved):
:+1:

Issue #273, opened 2022-05-20
Fails to load ParquetSharpNative.so on CentOS/RedHat


mfkl:
ParquetSharp fails to load its native dependencies on certain Linux distributions. The issue probably comes from a wrong nuget packaging configuration, as `ParquetSharpNative.so` is not copied to the `bin` folder like it should.

Stacktrace:
```
Error Message:
   System.TypeInitializationException : The type initializer for 'ParquetSharp.LogicalTypeFactory' threw an exception.
  ----> System.DllNotFoundException : Unable to load shared library 'ParquetSharpNative' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: libParquetSharpNative: cannot open shared object file: No such file or directory
  Stack Trace:
     at ParquetSharp.ParquetFileWriter..ctor(String path, Column[] columns, Compression compression, IReadOnlyDictionary`2 keyValueMetadata)
   at RQ.ParquetSharp.ParquetFileBuilder.WriteToParquet(String filename) in /Parquet/RQ.ParquetSharp/ParquetFileBuilder.cs:line 33
   at RQ.ParquetSharp.Test.TestParquetFileBuilder.TestParquetFileRoundTrip() in /Parquet/RQ.ParquetSharp.Test/TestParquetFileBuilder.cs:line 44
--DllNotFoundException
   at ParquetSharp.LogicalType.LogicalType_None(IntPtr& logicalType)
   at ParquetSharp.ExceptionInfo.Return[TValue](GetAction`1 getter)
   at ParquetSharp.LogicalType.None()
   at ParquetSharp.LogicalTypeFactory..cctor()
```

Affected Linux distributions:
- CentOS 7, 7.9.2009
- RedHat Enterprise 7.9

Affected ParquetSharp versions:
- Latest 7.0.0 Beta 1 on net5.0 (and older versions).

Debian 10 can load the native binary just fine.

jgiannuzzi:
The issue comes from a difference in versioned symbols in the system libraries between Ubuntu 18.04 (which we use to build ParquetSharp), and CentOS/RHEL 7. I'm investigating further what our options are to solve this. Off the top of my head, it may be possible to install an updated version of the C and C++ libraries side-by-side on CentOS/RHEL, or we could try building on an older version of Ubuntu (which would require building in a container as GitHub Actions does not support anything older than 18.04). Another idea would be to build a statically linked version of the native library.

dgl:
[...]

PR #274, opened 2022-05-24
Build linux-x64 in a CentOS 7 container for backwards compatibility

jgiannuzzi:
This PR aims at fixing #273 by building the linux-x64 native library within a CentOS 7 container with GCC 7. This allows linking with older versions of the system libraries, widening the Linux distributions supported by our NuGet package.

Please note that the linux-arm64 native library is still built on Ubuntu 18.04, as there is no recent cross-compiler available on CentOS 7. We will probably need to first get self-hosted linux-arm64 runners before we can support this. All of this being said, I don't expect many people to use old Linux distributions on arm64.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,17 @@ jobs:
           vcpkg-${{ steps.vcpkg-info.outputs.triplet }}-cmake${{ steps.cmake-info.outputs.version }}-
           vcpkg-${{ steps.vcpkg-info.outputs.triplet }}-
 
+    # Setup a CentOS 7 container to build on Linux x64 for backwards compatibility.
+    - name: Start CentOS container and install toolchain
+      if: runner.os == 'Linux' && matrix.arch == 'x64'
+      run: |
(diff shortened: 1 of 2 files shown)


adamreeve (approved):
Looks good thanks Jonathan :+1: 

It would be nice if we could cache the centos image with the build dependencies installed but that might be fairly complicated and it doesn't look like the container setup step takes too long

jgiannuzzi:
Indeed it would be a bit more complex and with no real advantage as it's already faster than the Windows build, which is usually the last to finish before we can move on with the workflow.

marcin-krystianc:
I think there is one small issue here with vcpkg caching and the cmake version. We will use runner's cmake version as part of the cache key, but the actual compilation occurs on the CentOS which uses different cmake.

jgiannuzzi:
@marcin-krystianc we install the same version of cmake in the container as on the host for that exact reason

marcin-krystianc:
> @marcin-krystianc we install the same version of cmake in the container as on the host for that exact reason

Ah, indeed. I've missed that. Perfect!

marcin-krystianc (approved):

PR #277, opened 2022-05-30
Improve error message when getting a LogicalColumnReader or Writer with the wrong type

adamreeve:
This is a pretty common error for users to encounter and the current error message looks like this: ```Unable to cast object of type 'ParquetSharp.LogicalColumnReader`3[System.Int32,System.Nullable`1[System.Int32],System.Nullable`1[System.Int32]]' to type 'ParquetSharp.LogicalColumnReader`1[System.Int32]'```. Unless you're familiar with the ParquetSharp code or have run into this before it's not very obvious what has gone wrong.

### DIFF
--- a/csharp.test/TestParquetFileReader.cs
+++ b/csharp.test/TestParquetFileReader.cs
@@ -55,7 +55,7 @@ namespace ParquetSharp.Test
                 }
             });
 
-            StringAssert.StartsWith("Unable to cast object of type", exception?.Message);
+            StringAssert.StartsWith("Tried to get a LogicalColumnReader", exception?.Message);
         }
 
         [Test]
--- a/csharp.test/TestLogicalColumnReader.cs
+++ b/csharp.test/TestLogicalColumnReader.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using System.Linq;
+using ParquetSharp.IO;
+using NUnit.Framework;
+
+namespace ParquetSharp.Test
+{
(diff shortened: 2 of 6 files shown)


csharp.test/TestParquetFileReader.cs
```diff
@@ -55,7 +55,7 @@ public static void TestFileHandleHasBeenReleased()
                 }
             });
 
-            StringAssert.StartsWith("Unable to cast object of type", exception?.Message);
+            StringAssert.StartsWith("Tried to get a LogicalColumnReader", exception?.Message);
```
marcin-krystianc, line 58: nit: NUnit has also fluent assertion syntax, I find it much nicer, e.g.:
```
Assert.That(exception?.Message, Does.StartWith("Tried to get a LogicalColumnReader"));
```
On the other hand, consistency is also important so it is probably good as it is now.

marcin-krystianc (dismissed):
Looks good 👍

csharp.test/TestParquetFileReader.cs
adamreeve: Good point. Most new tests are being written with the fluent assertion syntax so consistency is already gone, so I've taken this opportunity to update all the uses of `StringAssert`. I've also added the column name to the message too.

marcin-krystianc (approved):

jgiannuzzi (approved):

PR #279, opened 2022-06-01
Allow updating key-value metadata after ParquetFileWriter creation

adamreeve:
This PR changes ParquetFileWriter so that it doesn't read the KeyValueMetadata dictionary until the file is closed, to allow metadata entries to be added or updated at any time before the file is closed.

Internally, the Arrow C++ ParquetFileWriter takes a `std::shared_ptr<const KeyValueMetadata>` and then doesn't read the entries from the metadata until the file is closed and the footer is written, so this behaviour closely maps to the C++ API.

This changes behaviour and has the potential to be a breaking change if anyone is relying on being able to modify the metadata dictionary after passing it to the ParquetFileWriter constructor without this affecting the written metadata. That seems fairly unlikely though.

### DIFF
--- a/cpp/KeyValueMetadata.cpp
+++ b/cpp/KeyValueMetadata.cpp
@@ -9,7 +9,7 @@ using namespace parquet;
 
 extern "C"
 {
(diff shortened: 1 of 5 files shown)


marcin-krystianc (approved):
Looks good. I think you've covered all possible cases in the newly added tests - well done 👍

PR #280, opened 2022-06-03
Use fixed culture when testing OverflowException

Nisden:
I have changed the TestScaleOverflow test to use an fixed culture (en-US), as otherwise the test may fail depending on the users local culture.

An example is Danish where "," is the decimal separator instead of "."


jgiannuzzi (approved):
Good catch! 👍
Thanks a lot for your contribution @Nisden!

PR #281, opened 2022-06-03
Added support for force_set_converted_type

Nisden:
I have attempted to add support for `force_set_converted_type`, but I am a stuck as its not throwing any errors, its just ignoring my attempts.

But basically `force_set_converted_type` forces Parquet++ to write an "ConvertedType" for `Timestamp(isAdjustedToUTC=false)`, as otherwise its only written for `isAdjustedToUTC=true` and some legacy readers only support "ConvertedType" and not the newer "LogicalType"

Relevant commits are:
  * https://github.com/apache/arrow/commit/9189e0083120689f5d76291024d165f69279d03f
  * https://github.com/apache/arrow/commit/99b9702a76331a572e841c9a6d1eae9f725e9616


Nisden:
@jgiannuzzi Any chance you can provide some assistance here? I have attempted to solve this but I am stuck.
What happens right now is that `force_set_converted_type` seems to be set (According to the unit test), but when I read back in the file `is_from_converted_type` is not set.

Also if I try and write out the file and open it with parquet-tools the inspect command confirms that file was NOT written with `force_set_converted_type` nor a `converted_type` 

```
############ Column(a) ############
name: a
path: a
max_definition_level: 0
max_repetition_level: 0
physical_type: INT64
logical_type: Timestamp(isAdjustedToUTC=false, timeUnit=milliseconds, is_from_converted_type=false, force_set_converted_type=false)
converted_type (legacy): NONE
```

adamreeve:
Hi @Nisden. I've had a look into this and I believe your change is working as expected. Have you tested reading data with one of the legacy readers that only supports ConvertedType?

When you read back the `ForceSetConvertedType` property in your test you're expecting it to be true but it's actually false. This is because the `ForceSetConvertedType` isn't persisted into the Parquet file data. It's only used when writing files and is just kept in memory to control the writing behaviour. So when you read column metadata from a file you've opened for reading it will always be false. (I'm not super familiar with this so could be wrong, but this is my understanding of what's going on)

The `parquet-tools` output is a little surprising, I'm not sure what's happening there but I suspect that Arrow doesn't set the converted type if a logical type is present.

I tested modifying your `TestForceSetConvertedTypeRoundtrip` test to write out a file to disk and then read the Parquet footer Thrift data, and it looks like the converted type is definitely being set in the file metadata. The column schema element looks like:
```
SchemaElement(
  type=2,
  type_length=None,
  repetition_type=0,
  name='a',
  num_children=None,
  converted_type=9,
  scale=None,
  precision=None,
  field_id=None,
  logicalType=LogicalType(
    STRING=None,
    MAP=None,
    LIST=None,
    ENUM=None,
    DECIMAL=None,
    DATE=None,
    TIME=None,
    TIMESTAMP=TimestampType(isAdjustedToUTC=False, unit=TimeUnit(MILLIS=MilliSeconds(), MICROS=None, NANOS=None)),
    INTEGER=None,
    UNKNOWN=None,
    JSON=None,
    BSON=None,
    UUID=None))
```
where `converted_type=9` corresponds to `TIMESTAMP_MILLIS`. I also checked the result when writing with `forceSetConvertedType: false`, which gives `converted_type=None`.

For reference, here's how I read the footer metadata:

Generate Python code for reading Thrift data:
```
curl -o parquet.thrift https://raw.githubusercontent.com/apache/parquet-format/master/src/main/thrift/parquet.thrift
thrift -r --gen py parquet.thrift
```

Then read the metadata using a Python script:
```Python
import struct
from thrift.transport import TTransport
from thrift.protocol import TCompactProtocol
import sys
sys.path.append('./gen-py')
from parquet.ttypes import FileMetaData


with open(sys.argv[1], 'rb') as f:
    data = f.read()

footer_len = data[-8:-4]
footer_len, = struct.unpack('<l', footer_len)
footer = data[-footer_len - 8:-8]

transport = TTransport.TMemoryBuffer(footer)
protocol = TCompactProtocol.TCompactProtocol(transport)

metadata = FileMetaData()
metadata.read(protocol)

schema = metadata.schema
col = schema[1]

print(col)
```

Nisden:
Hi @adamreeve 

Thanks for your detailed look into this! 
Ill try and confirm your findings as well, and then we can look at merging this.

I personally got a bit thrown off by parquet-tools but also because Synapse Serverless still refuse to query our files and I believed it was because it was missing ConvertedType.

Nisden:
@adamreeve I confirmed you result using https://github.com/Eugene-Mark/bigdata-file-viewer, it uses an old Java version for reading Parquet and correctly displayed that no ConvertedType was set when ForceSetConvertedType was missing.

I removed my unit test as it provided no real value, and I am unsure how a unit test for this feature should be structured when Parquet-cpp seems unable to read ConvertedType.

Any suggestions on how a test could look?

adamreeve:
Good question, I think it's worth having a test case that exercises the code path where `forceSetConvertedType` is true, just to check it doesn't cause a crash, but it might not be worth testing that the converted type is set if it's too complicated.

It might be possible to expose the converted type so you can access it in your test with `columnReader.ColumnDescriptor.SchemaNode.ConvertedType` for example, although possibly that won't be set if the C++ Arrow library behaviour matches PyArrow.

Nisden:
I have implemented `ConvertedType` but sadly your correct that its only set when writing and not when reading.
However it still allows for a better test case, as we can check that `forceSetConvertedType` makes a difference to `ConvertedType`

csharp/Schema/ConvertedType.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        None = 0,  // Not a real converted type, but means no converted type is specified
+        UTF8 = 1,
+        Map = 2,
+        MapKeyValue = 3,
+        List = 4,
+        Enum = 5,
+        Decimal = 6,
+        Date = 7,
+        Time_Millis = 8,
```
adamreeve, line 13: Can all of the enum members please be named without underscores

adamreeve (changes requested):
Looks good to me thanks Nick! There are just some minor formatting issues to fix. You can run the automatic formatter to make sure the format check passes with:
```
dotnet tool restore
dotnet jb cleanupcode "csharp" "csharp.test" "csharp.benchmark" --profile="Built-in: Reformat Code" --settings="ParquetSharp.DotSettings"
```
(the enum member naming will need manual tidying though)

adamreeve (dismissed):

csharp/LogicalType.cs
```diff
@@ -59,7 +59,7 @@ public override string ToString()
         public static LogicalType Decimal(int precision, int scale = 0) => Create(ExceptionInfo.Return<int, int, IntPtr>(precision, scale, LogicalType_Decimal));
         public static LogicalType Date() => Create(ExceptionInfo.Return<IntPtr>(LogicalType_Date));
         public static LogicalType Time(bool isAdjustedToUtc, TimeUnit timeUnit) => Create(ExceptionInfo.Return<bool, TimeUnit, IntPtr>(isAdjustedToUtc, timeUnit, LogicalType_Time));
-        public static LogicalType Timestamp(bool isAdjustedToUtc, TimeUnit timeUnit) => Create(ExceptionInfo.Return<bool, TimeUnit, IntPtr>(isAdjustedToUtc, timeUnit, LogicalType_Timestamp));
+        public static LogicalType Timestamp(bool isAdjustedToUtc, TimeUnit timeUnit, bool forceSetConvertedType = false) => Create(ExceptionInfo.Return<bool, TimeUnit, bool, IntPtr>(isAdjustedToUtc, timeUnit, forceSetConvertedType, LogicalType_Timestamp));
```
adamreeve, line 62:
```suggestion
        public static LogicalType Timestamp(bool isAdjustedToUtc, TimeUnit timeUnit) => Create(ExceptionInfo.Return<bool, TimeUnit, bool, IntPtr>(isAdjustedToUtc, timeUnit, false, LogicalType_Timestamp));
        public static LogicalType Timestamp(bool isAdjustedToUtc, TimeUnit timeUnit, bool forceSetConvertedType) => Create(ExceptionInfo.Return<bool, TimeUnit, bool, IntPtr>(isAdjustedToUtc, timeUnit, forceSetConvertedType, LogicalType_Timestamp));
```

Hi @Nisden, apologies for asking for another change, but it would be good to avoid changing the ABI here by using an overload instead of adding an optional parameter.

We're working on getting the 8.0.0 release out now so will aim to include this PR.

adamreeve (approved):

PR #284, opened 2022-06-13
Allow specifying column definitions for a ParquetRowWriter

adamreeve:
This gives users more control over the logical type mappings for columns, eg. to specify `IsAdjustedToUtc` for Timestamps, and works when using a tuple type where we can't add attributes to fields.

A column definition specified will take precedence over any attributes present on a type (`MapToColumn` or `ParquetDecimalScale`).


marcin-krystianc (dismissed):
Looks good.

jgiannuzzi (approved):

PR #285, opened 2022-06-14
Add support for .NET 6

jgiannuzzi:
This PR adds support for .NET 6, both in project files and in the CI.

Summary of the changes:
- require .NET 6 SDK to build via `global.json`
- add `net6.0` to test target frameworks, and only build for it by default
- switch target framework to `net6.0` for benchmarks
- use `HandleProcessCorruptedStateExceptions` on .NET Framework only (it never worked on .NET Core and is unavailable on .NET 6)
- always use the latest 6.0.x SDK instead of a pinned version in CI
- daily CI runs of the CI to get notified of potential build failures when a new version of the SDK is released
- CI runs on both macOS 10.15 and 11.0

### DIFF
--- a/csharp.test/ParquetSharp.Test.csproj
+++ b/csharp.test/ParquetSharp.Test.csproj
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)'=='Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(GITHUB_ACTIONS)' == 'true'">$(TargetFrameworks);netcoreapp3.1;net5.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(GITHUB_ACTIONS)' == 'true' AND '$(OS)'=='Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
     <LangVersion>9.0</LangVersion>
     <Nullable>enable</Nullable>
     <AssemblyName>ParquetSharp.Test</AssemblyName>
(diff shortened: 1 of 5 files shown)


csharp.test/ParquetSharp.Test.csproj
```diff
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)'=='Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(GITHUB_ACTIONS)' == 'true'">$(TargetFrameworks);netcoreapp3.1;net5.0</TargetFrameworks>
```
marcin-krystianc, line 5: I think that we should avoid unnecessary discrepancy between local testing and CI. Also, there is no benefit from removing other target frameworks for local development and testing, or is it?

jgiannuzzi: This is just to avoid getting errors when running the tests locally on a machine that does not have the required runtimes installed. It also avoids building those target frameworks by default, speeding up the build process. And then our CI ensures that those frameworks still build and run without warnings or errors.

FYI I was inspired by what we started doing with [ILGPU](https://github.com/m4rs-mt/ILGPU/blob/v1.2.0/Src/Directory.Build.props).

jgiannuzzi: Happy to revert this if we don't feel comfortable with it!

marcin-krystianc: > Happy to revert this if we don't feel comfortable with it!

Nah, I think it is ok. I would just add some comment explaining our motivation. Also, I think that `Condition="'$(CI)' == 'true'">` will be better (as it is more universal).

jgiannuzzi: Great point!

marcin-krystianc (approved):

PR #288, opened 2022-06-23
Added switch to force DateTimeKind.Unspecified

Nisden:
This pull request add's an AppContext switch to enable the old behavior of always reading `DateTime` as `DateTimeKind.Unspecified`

To use this you can simply run `AppContext.SetSwitch("ParquetSharp.UseDateTimeKindUnspecified", true);` or use `RuntimeHostConfigurationOption` in your .csproj file.

The reason behind this, is that we have software in our stack that forces us to set `isAdjustedToUtc=true` even though the values are actually local time.

### DIFF
--- a/csharp.test/TestLogicalTypeRoundtrip.cs
+++ b/csharp.test/TestLogicalTypeRoundtrip.cs
@@ -150,8 +150,8 @@ namespace ParquetSharp.Test
             // This test cannot be parallelized as we use an AppContext switch to manipulate the internal behavior of ParquetSharp.
             // If other test's run while this test is also running it may cause inconsistent results.
 
-            Assert.False(AppContext.TryGetSwitch("ParquetSharp.UseDateTimeKindUnspecified", out var existingValue) && existingValue);
-            AppContext.SetSwitch("ParquetSharp.UseDateTimeKindUnspecified", true);
+            Assert.False(AppContext.TryGetSwitch("ParquetSharp.ReadDateTimeKindAsUnspecified", out var existingValue) && existingValue);
+            AppContext.SetSwitch("ParquetSharp.ReadDateTimeKindAsUnspecified", true);
 
             try
             {
@@ -205,7 +205,7 @@ namespace ParquetSharp.Test
             }
             finally
             {
-                AppContext.SetSwitch("ParquetSharp.UseDateTimeKindUnspecified", false);
+                AppContext.SetSwitch("ParquetSharp.ReadDateTimeKindAsUnspecified", false);
             }
         }
 
--- a/csharp.test/TestLogicalTypeRoundtrip.cs
+++ b/csharp.test/TestLogicalTypeRoundtrip.cs
@@ -665,6 +665,78 @@ namespace ParquetSharp.Test
             Assert.That(values, Is.EqualTo(expected));
         }
 
+        [Test]
+        public static void TestForceSetConvertedTypeSetsConvertedType()
+        {
+            var expected = new DateTime[]
+            {
+                new DateTime(2000, 1, 1)
+            };
+
+            using var buffer = new ResizableBuffer();
+
+            using (var outStream = new BufferOutputStream(buffer))
+            {
+                using var fileWriter = new ParquetFileWriter(outStream, new Column[] {new Column<DateTime>("a", LogicalType.Timestamp(false, TimeUnit.Millis, forceSetConvertedType: true))});
+                using var rowGroupWriter = fileWriter.AppendRowGroup();
+                using var colWriter = rowGroupWriter.NextColumn().LogicalWriter<DateTime>();
+
+                Assert.True((colWriter.ColumnDescriptor.LogicalType as TimestampLogicalType)?.ForceSetConvertedType);
+                Assert.AreEqual(ConvertedType.TimestampMillis, colWriter.ColumnDescriptor.SchemaNode.ConvertedType);
+
+                colWriter.WriteBatch(expected);
+
+                fileWriter.Close();
+            }
+
+            using var inStream = new BufferReader(buffer);
+            using var fileReader = new ParquetFileReader(inStream);
+            using var rowGroup = fileReader.RowGroup(0);
+            using var columnReader = rowGroup.Column(0).LogicalReader<DateTime>();
+
+            Assert.AreEqual(1, rowGroup.MetaData.NumRows);
+            var allData = columnReader.ReadAll(1);
+            Assert.AreEqual(expected, allData);
+        }
+
+        [Test]
+        public static void TestTimestampLocalHasNoConvertedType()
+        {
+            // This test works in tandem with TestForceSetConvertedTypeSetsConvertedType
+            // They confirm that LogicalType.Timestamp(false) has no ConvertedType by default, but when you set forceSetConvertedType: true the ConvertedType is set.
+            // However it seems like we can only confirm this when writing, and reading ignores ConvertedType
+
+            var expected = new DateTime[]
+            {
+                new DateTime(2000, 1, 1)
+            };
+
+            using var buffer = new ResizableBuffer();
+
+            using (var outStream = new BufferOutputStream(buffer))
+            {
+                using var fileWriter = new ParquetFileWriter(outStream, new Column[] {new Column<DateTime>("a", LogicalType.Timestamp(false, TimeUnit.Millis))});
+                using var rowGroupWriter = fileWriter.AppendRowGroup();
+                using var colWriter = rowGroupWriter.NextColumn().LogicalWriter<DateTime>();
+
+                Assert.False((colWriter.ColumnDescriptor.LogicalType as TimestampLogicalType)?.ForceSetConvertedType);
+                Assert.AreEqual(ConvertedType.None, colWriter.ColumnDescriptor.SchemaNode.ConvertedType);
+
+                colWriter.WriteBatch(expected);
+
+                fileWriter.Close();
+            }
+
+            using var inStream = new BufferReader(buffer);
+            using var fileReader = new ParquetFileReader(inStream);
+            using var rowGroup = fileReader.RowGroup(0);
+            using var columnReader = rowGroup.Column(0).LogicalReader<DateTime>();
+
+            Assert.AreEqual(1, rowGroup.MetaData.NumRows);
+            var allData = columnReader.ReadAll(1);
+            Assert.AreEqual(expected, allData);
+        }
+
         private static ExpectedColumn[] CreateExpectedColumns()
         {
             return new[]
--- a/csharp.test/TestLogicalTypeRoundtrip.cs
+++ b/csharp.test/TestLogicalTypeRoundtrip.cs
@@ -143,6 +143,68 @@ namespace ParquetSharp.Test
             Assert.AreEqual(kind, kinds.First());
         }
 
+        [Test]
+        [NonParallelizable]
+        public static void TestAppSwitchDateTimeKindUnspecified()
+        {
+            AppContext.SetSwitch("ParquetSharp.UseDateTimeKindUnspecified", true);
+
+            try
+            {
+                // ParquetSharp doesn't know the DateTime values upfront,
+                // so we have to specify whether values are UTC in the logical type.
+                var schemaColumns = new Column[]
+                {
+                    new Column<DateTime>("a", LogicalType.Timestamp(true, TimeUnit.Millis)),
+                    new Column<DateTime>("b", LogicalType.Timestamp(false, TimeUnit.Millis)),
+                };
+
+                const int numRows = 100;
+                var startTime = new DateTime(2022, 3, 14, 10, 49, 0, DateTimeKind.Unspecified);
+                var values = Enumerable.Range(0, numRows).Select(i => startTime + TimeSpan.FromSeconds(i)).ToArray();
+
+                using var buffer = new ResizableBuffer();
+
+                using (var outStream = new BufferOutputStream(buffer))
+                {
+                    using var fileWriter = new ParquetFileWriter(outStream, schemaColumns);
+                    using var rowGroupWriter = fileWriter.AppendBufferedRowGroup();
+                    using var columnWriterA = rowGroupWriter.Column(0).LogicalWriter<DateTime>();
+                    columnWriterA.WriteBatch(values);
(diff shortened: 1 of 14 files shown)


adamreeve:
Hi @Nisden, have you considered using a custom `LogicalReadConverterFactory` to handle this? Something like this should achieve what you want:

```C#
class UnspecifiedDateTimeConverterFactory : LogicalReadConverterFactory
{
    public override Delegate GetConverter<TLogical, TPhysical>(ColumnDescriptor columnDescriptor,
        ColumnChunkMetaData columnChunkMetaData)
    {
        if (typeof(TLogical) == typeof(DateTime))
        {
            var timestampType = (TimestampLogicalType) columnDescriptor.LogicalType;
            switch (timestampType.TimeUnit)
            {
                case TimeUnit.Millis:
                    return (LogicalRead<DateTime, long>.Converter) ((s, _, d, _) => LogicalRead.ConvertDateTimeMillis(s, d, DateTimeKind.Unspecified));
                case TimeUnit.Micros:
                    return (LogicalRead<DateTime, long>.Converter) ((s, _, d, _) => LogicalRead.ConvertDateTimeMicros(s, d, DateTimeKind.Unspecified));
            }
        }
        if (typeof(TLogical) == typeof(DateTime?))
        {
            var timestampType = (TimestampLogicalType) columnDescriptor.LogicalType;
            switch (timestampType.TimeUnit)
            {
                case TimeUnit.Millis:
                    return (LogicalRead<DateTime?, long>.Converter) (
                        (source, levels, dest, def) => LogicalRead.ConvertDateTimeMillis(source, levels, dest, def, DateTimeKind.Unspecified));
                case TimeUnit.Micros:
                    return (LogicalRead<DateTime?, long>.Converter) (
                        (source, levels, dest, def) => LogicalRead.ConvertDateTimeMicros(source, levels, dest, def, DateTimeKind.Unspecified));
            }
        }
        return base.GetConverter<TLogical, TPhysical>(columnDescriptor, columnChunkMetaData);
    }
}
```

Which you can then use with a `ParquetFileReader`:
```C#
using var fileReader = new ParquetFileReader(filename)
{
    LogicalReadConverterFactory = new UnspecifiedDateTimeConverterFactory()
};
```

Nisden:
@adamreeve Sure that could work, but I just found it cleaner to make it an AppSwitch, as I can enforce that globally. (And we did talk about a way to disable it back in https://github.com/G-Research/ParquetSharp/issues/260)

adamreeve:
Right, sorry I forgot you mentioned you were already using a custom LogicalReadConverterFactory. Yes this seems like a nice approach to allow keeping compatibility with the old behavior without needing to change code. Would you mind adding a section to the documentation to document this new switch? Probably something in docs/Reading.md would make sense.

csharp.test/TestLogicalTypeRoundtrip.cs
```diff
@@ -143,6 +143,68 @@ public static void TestDateTimeRoundTrip(DateTimeKind kind, TimeUnit timeUnit)
             Assert.AreEqual(kind, kinds.First());
         }
 
+        [Test]
+        [NonParallelizable]
+        public static void TestAppSwitchDateTimeKindUnspecified()
+        {
+            AppContext.SetSwitch("ParquetSharp.UseDateTimeKindUnspecified", true);
```
marcin-krystianc, line 150: I think that it would be good to assert that the switch is not set before we start manipulating it. The precondition of this test that all other tests expect the switch to be false. In case someone uses registry entry to set a switch he would get weird tests errors, so such assertion would help to debug such issue.

csharp.test/TestLogicalTypeRoundtrip.cs
```diff
@@ -143,6 +143,68 @@ public static void TestDateTimeRoundTrip(DateTimeKind kind, TimeUnit timeUnit)
             Assert.AreEqual(kind, kinds.First());
         }
 
+        [Test]
+        [NonParallelizable]
```
marcin-krystianc, line 147: I think it deserves a comment why this test shouldn't run together with other tests.

csharp.test/TestLogicalTypeRoundtrip.cs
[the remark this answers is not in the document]
Nisden: Would that not be bad practice, as you start exposing implementation details where its otherwise unnecessary?

Also this is inline with how ASP.NET Core does it: `{namespace}.{effect}`:
  * https://github.com/dotnet/aspnetcore/blob/5026aefddd69322da633d4b29a3d24c4df0b51b9/src/Http/Http/src/Internal/ResponseCookies.cs#L17
  * https://github.com/dotnet/aspnetcore/blob/84c0baa9a57bd6448cd1bf78b64d2c1658bf5ef6/src/Identity/Core/src/SignInManager.cs#L387

docs/Reading.md
```diff
@@ -97,3 +97,23 @@ while (logicalColumnReader.HasNext)
 
 The .NET type used to represent read values can optionally be overridden by using the `ColumnReader.LogicalReaderOverride<TElement>` method.
 For more details, see the [type factories documentation](TypeFactories.md).
+
+## DateTimeKind when reading Timestamp's
+When reading Timestamp's to an DateTime, ParquetSharp will set the DateTimeKind based on the value of `IsAdjustedToUtc`.
```
marcin-krystianc, line 102:
```suggestion
When reading Timestamp to a DateTime, ParquetSharp sets the DateTimeKind based on the value of `IsAdjustedToUtc`.
```

docs/Reading.md
```diff
@@ -97,3 +97,23 @@ while (logicalColumnReader.HasNext)
 
 The .NET type used to represent read values can optionally be overridden by using the `ColumnReader.LogicalReaderOverride<TElement>` method.
 For more details, see the [type factories documentation](TypeFactories.md).
+
+## DateTimeKind when reading Timestamp's
```
marcin-krystianc, line 101:
```suggestion
## DateTimeKind when reading Timestamps
```

docs/Reading.md
```diff
@@ -97,3 +97,23 @@ while (logicalColumnReader.HasNext)
...
+When reading Timestamp's to an DateTime, ParquetSharp will set the DateTimeKind based on the value of `IsAdjustedToUtc`.
+
+If `IsAdjustedToUtc` is `true` the DateTimeKind will be set to `DateTimeKind.Utc` otherwise it will be set to `DateTimeKind.Unspecified`.
+
+This behavior can be overwritten by setting the AppContext switch `ParquetSharp.UseDateTimeKindUnspecified` to `true`.  
+This also matches the old behavior of [ParquetSharp > 6.0.0](https://github.com/G-Research/ParquetSharp/pull/261)
```
marcin-krystianc, line 107: Shouldn't it be less than version 7.0.0?
```suggestion
This also matches the old behavior of [ParquetSharp < 7.0.0](https://github.com/G-Research/ParquetSharp/pull/261)
```

docs/Reading.md
```diff
@@ -97,3 +97,23 @@ while (logicalColumnReader.HasNext)
 
 The .NET type used to represent read values can optionally be overridden by using the `ColumnReader.LogicalReaderOverride<TElement>` method.
 For more details, see the [type factories documentation](TypeFactories.md).
+
+## DateTimeKind when reading Timestamp's
+When reading Timestamp's to an DateTime, ParquetSharp will set the DateTimeKind based on the value of `IsAdjustedToUtc`.
+
+If `IsAdjustedToUtc` is `true` the DateTimeKind will be set to `DateTimeKind.Utc` otherwise it will be set to `DateTimeKind.Unspecified`.
+
+This behavior can be overwritten by setting the AppContext switch `ParquetSharp.UseDateTimeKindUnspecified` to `true`.
```
marcin-krystianc, line 106:
```suggestion
This behavior can be overwritten by setting the AppContext switch `ParquetSharp.UseDateTimeKindUnspecified` to `true`, so the DateTimeKind will be always set to `DateTimeKind.Unspecified` regardless of the value of `IsAdjustedToUtc`.
```

csharp/LogicalRead.cs
```diff
@@ -10,6 +10,8 @@ namespace ParquetSharp
     public static class LogicalRead<TLogical, TPhysical>
         where TPhysical : unmanaged
     {
+        private const string UseDateTimeKindUnspecifiedSwitchName = "ParquetSharp.UseDateTimeKindUnspecified";
```
marcin-krystianc, line 13: You can make it `internal` instead of `private` so it can be used in tests as well.

marcin-krystianc (commented):
@Nisden Thanks for your changes, I've left few more comments.

docs/Reading.md
Nisden: Your right! I messed up the sign and [version](https://github.com/G-Research/ParquetSharp/releases/tag/7.0.0).

csharp/LogicalRead.cs
Nisden: Personally not a fan of that because:
  1. You would have to access the variable with generics specified `LogicalRead<DateTime, long>.UseDateTimeKindUnspecifiedSwitchName`
  2. You will no longer be able to look at a unit test as a reference for how it can be implemented.

csharp.test/TestLogicalTypeRoundtrip.cs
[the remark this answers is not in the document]
Nisden: I would be fine with `ParquetSharp.ReadDateTimeKindAsUnspecified`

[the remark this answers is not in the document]
adamreeve: 👍 `ParquetSharp.ReadDateTimeKindAsUnspecified` sounds good to me

csharp/LogicalRead.cs
marcin-krystianc: There is static non-generic `LogicalRead` class that could be used for that, but I agree there are some befits of using raw strings in the test code. I don't have strong opinion here, I'll leave it to your judgment.

adamreeve (dismissed):

marcin-krystianc (commented):
@Nisden 
I've just realised that we didn't cover `DateTime?` in tests.
Can you update the test method with this code?:
```
        [Test]
        [NonParallelizable]
        public static void TestAppSwitchDateTimeKindUnspecified()
        {
            // This test cannot be parallelized as we use an AppContext switch to manipulate the internal behavior of ParquetSharp.
            // If other test's run while this test is also running it may cause inconsistent results.

            Assert.False(AppContext.TryGetSwitch("ParquetSharp.ReadDateTimeKindAsUnspecified", out var existingValue) && existingValue);
            AppContext.SetSwitch("ParquetSharp.ReadDateTimeKindAsUnspecified", true);

            try
            {
                // We create two Timestamp columns with varying isAdjustedToUtc
                // With the legacy switch enabled, both columns when read should output DateTime values with DateTimeKind.Unspecified
                var schemaColumns = new Column[]
                {
                    new Column<DateTime>("a", LogicalType.Timestamp(true, TimeUnit.Millis)),
                    new Column<DateTime>("b", LogicalType.Timestamp(false, TimeUnit.Millis)),
                    new Column<DateTime?>("c", LogicalType.Timestamp(true, TimeUnit.Millis)),
                    new Column<DateTime?>("d", LogicalType.Timestamp(false, TimeUnit.Millis)),
                };

                const int numRows = 100;
                var startTime = new DateTime(2022, 3, 14, 10, 49, 0, DateTimeKind.Unspecified);
                var values = Enumerable.Range(0, numRows).Select(i => startTime + TimeSpan.FromSeconds(i)).ToArray();

                using (var outStream = new BufferOutputStream(buffer))
                {
                    using var fileWriter = new ParquetFileWriter(outStream, schemaColumns);
                    using var rowGroupWriter = fileWriter.AppendBufferedRowGroup();
                    using var columnWriterA = rowGroupWriter.Column(0).LogicalWriter<DateTime>();
                    columnWriterA.WriteBatch(values);

                    using var columnWriterB = rowGroupWriter.Column(1).LogicalWriter<DateTime>();
                    columnWriterB.WriteBatch(values);

                    using var columnWriterC = rowGroupWriter.Column(2).LogicalWriter<DateTime?>();
                    columnWriterC.WriteBatch(values.Cast<DateTime?>().ToArray());

                    using var columnWriterD = rowGroupWriter.Column(3).LogicalWriter<DateTime?>();
                    columnWriterD.WriteBatch(values.Cast<DateTime?>().ToArray());

                    fileWriter.Close();
                }

                DateTime[] readValuesA;
                DateTime[] readValuesB;
                DateTime?[] readValuesC;
                DateTime?[] readValuesD;
                using (var inStream = new BufferReader(buffer))
                {
                    using var fileReader = new ParquetFileReader(inStream);
                    using var rowGroupReader = fileReader.RowGroup(0);
                    using var columnReaderA = rowGroupReader.Column(0);
                    using var logicalReaderA = columnReaderA.LogicalReader<DateTime>();
                    readValuesA = logicalReaderA.ReadAll(numRows);

                    using var columnReaderB = rowGroupReader.Column(1);
                    using var logicalReaderB = columnReaderB.LogicalReader<DateTime>();
                    readValuesB = logicalReaderB.ReadAll(numRows);

                    using var columnReaderC = rowGroupReader.Column(2);
                    using var logicalReaderC = columnReaderC.LogicalReader<DateTime?>();
                    readValuesC = logicalReaderC.ReadAll(numRows);

                    using var columnReaderD = rowGroupReader.Column(3);
                    using var logicalReaderD = columnReaderD.LogicalReader<DateTime?>();
                    readValuesD = logicalReaderD.ReadAll(numRows);
                }

                Assert.AreEqual(values, readValuesA);
                Assert.AreEqual(values, readValuesB);
                Assert.AreEqual(values, readValuesC);
                Assert.AreEqual(values, readValuesD);

                var kinds = readValuesA.Select(v => v.Kind)
                    .Concat(readValuesB.Select(v => v.Kind))
                    .Concat(readValuesC.Select(v => v!.Value.Kind))
                    .Concat(readValuesD.Select(v => v!.Value.Kind))
                    .ToHashSet();
                Assert.AreEqual(1, kinds.Count);
                Assert.AreEqual(DateTimeKind.Unspecified, kinds.First());
            }
            finally
            {
                AppContext.SetSwitch("ParquetSharp.ReadDateTimeKindAsUnspecified", false);
            }
        }
```

Nisden:
Ill look at it, when I am back from vacation.

marcin-krystianc (approved):
Looks good, @Nisden thank you for your contribution and patience!

PR #289, opened 2022-06-23
Include runner's image identity into the cache key

### LINKED ISSUE #286: [CI] vcpkg caching seems broken for macOS
Our caching mechanisms for `vcpkg` packages seems to be broken on macOS.

Here is an example:
* `vcpkg` packages get [cached](https://github.com/G-Research/ParquetSharp/runs/6878386852?check_suite_focus=true#step:35:5)
* a subsequent workflow on the same branch [restores](https://github.com/G-Research/ParquetSharp/runs/6883652224?check_suite_focus=true#step:5:20) them
* the packages are [not used](https://github.com/G-Research/ParquetSharp/runs/6883652224?check_suite_focus=true#step:11:97) in that workflow and get rebuilt

marcin-krystianc:
Fixes #286 

There are many factors that can affect the binary identity of vcpkg's packages (https://vcpkg.readthedocs.io/en/stable/users/binarycaching/). Since we use vcpkg, cmake, and C/C+ compilers that are preinstalled on the runner, we don't have full control over which version of these tools exactly is being run. 
To make the action level caching consistent with the vcpkg binary cache, we need to include the runner's image identity in the cache key.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,17 +85,33 @@ jobs:
       run: echo "::set-output name=version::$(cmake --version | head -n1 | awk '{print $3}')"
       shell: bash
 
+    - name: Runner-info
+      id: runner-info
+      run: |
+        env
+        echo "::set-output name=Info::$ImageOS-$ImageVersion"
+      shell: bash
+
     # Check for cached vcpkg dependencies (use these if we can).
     - name: Get cached vcpkg dependencies
       id: get-cached-vcpkg
       uses: actions/cache@v3
       with:
         path: cache/vcpkg
-        key: vcpkg-${{ steps.vcpkg-info.outputs.triplet }}-cmake${{ steps.cmake-info.outputs.version }}-${{ hashFiles('vcpkg-configuration.json') }}-${{ hashFiles('vcpkg.json') }}
+        # https://vcpkg.readthedocs.io/en/stable/users/binarycaching/
+        # Binary caching relies on hashing everything that contributes to a particular package build. This includes:
+        # - The triplet file and name
+        # - The C and C++ compilers executable
+        # - The version of CMake used
+        # - Every file in the port directory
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -85,17 +85,33 @@ jobs:
       run: echo "::set-output name=version::$(cmake --version | head -n1 | awk '{print $3}')"
       shell: bash
 
+    - name: Runner-info
+      id: runner-info
+      run: |
+        env
+        echo "::set-output name=Info::$ImageOS-$ImageVersion"
```
marcin-krystianc, line 92: `Env` context works only for variables directly defined in the workflow. Therefore we need to use an extra step to be able to read the values of environment variables that are not defined directly in the workflow.

.github/workflows/ci.yml
```diff
@@ -87,17 +87,33 @@ jobs:
       run: echo "::set-output name=version::$(cmake --version | head -n1 | awk '{print $3}')"
       shell: bash
 
+    - name: Runner-info
+      id: runner-info
+      run: |
+        env
```
jgiannuzzi, line 93: I imagine this line came from debugging?
```suggestion
```

.github/workflows/ci.yml
jgiannuzzi: For consistency with the rest of the workflow, I'd suggest using a lowercase variable name.
```suggestion
        echo "::set-output name=info::$ImageOS-$ImageVersion"
```

jgiannuzzi (commented):
Great stuff, thanks @marcin-krystianc!

I have a few nitpicks below, but otherwise LGTM

marcin-krystianc:
@jgiannuzzi Thanks for your review, I've applied your suggestion. Please, can you have another look?

jgiannuzzi (approved):

PR #290, opened 2022-07-04
Dispose of most uses of LogicalType and Node

adamreeve:
In this PR I've tried to dispose of most internal uses of LogicalType and Node to reduce the need to run finalizers.

There is still one main place where we create logical types and nodes that aren't disposed that I'm aware of. In LogicalColumnStream, the LogicalType and SchemaNodesPath properties are initialized in the constructor rather than being getters that create new instances, so these should probably be disposed when disposing the LogicalColumnStream instance. But there's a small chance that could be a breaking change if somebody is expecting to be able to continue using those after the LogicalColumnStream is disposed, so I've currently left that as-is.

There's also `RowOriented.ParquetFile.GetColumn` which creates logical types that aren't disposed, and I couldn't see a straightforward way to fix that.

### DIFF
--- a/csharp.test/TestByteArrayReaderCache.cs
+++ b/csharp.test/TestByteArrayReaderCache.cs
@@ -29,7 +29,8 @@ namespace ParquetSharp.Test
             // Write a file that contains a lot of duplicate strings.
             using (var output = new BufferOutputStream(buffer))
             {
-                using var fileWriter = new ParquetFileWriter(output, columns, CreateWriterProperties(enableDictionary));
+                using var writerProperties = CreateWriterProperties(enableDictionary);
+                using var fileWriter = new ParquetFileWriter(output, columns, writerProperties);
                 using var groupWriter = fileWriter.AppendRowGroup();
 
                 using var dateWriter = groupWriter.NextColumn().LogicalWriter<DateTime>();
--- a/csharp/ColumnDescriptor.cs
+++ b/csharp/ColumnDescriptor.cs
@@ -69,11 +69,22 @@ namespace ParquetSharp
             var (physicalType, logicalType) = typeFactory.GetSystemTypes(this, columnLogicalTypeOverride);
             var elementType = logicalType;
 
-            for (var node = SchemaNode; node != null; node = node.Parent)
+            var node = SchemaNode;
+            while (node != null)
             {
-                if (node.LogicalType.Type == LogicalTypeEnum.List)
+                try
                 {
-                    elementType = elementType.MakeArrayType();
+                    using var nodeLogicalType = node.LogicalType;
+                    if (nodeLogicalType.Type == LogicalTypeEnum.List)
+                    {
+                        elementType = elementType.MakeArrayType();
+                    }
+                }
+                finally
+                {
+                    var prevNode = node;
+                    node = prevNode.Parent;
+                    prevNode.Dispose();
                 }
             }
 
--- a/csharp/ColumnDescriptor.cs
+++ b/csharp/ColumnDescriptor.cs
@@ -72,20 +72,15 @@ namespace ParquetSharp
             var node = SchemaNode;
             while (node != null)
             {
-                try
+                using var nodeLogicalType = node.LogicalType;
+                if (nodeLogicalType.Type == LogicalTypeEnum.List)
                 {
-                    using var nodeLogicalType = node.LogicalType;
-                    if (nodeLogicalType.Type == LogicalTypeEnum.List)
-                    {
-                        elementType = elementType.MakeArrayType();
-                    }
+                    elementType = elementType.MakeArrayType();
                 }
-                finally
-                {
-                    var prevNode = node;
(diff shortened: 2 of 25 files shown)


csharp/ParquetFileReader.cs
```diff
@@ -7,33 +7,37 @@ namespace ParquetSharp
...
         }
 
         public ParquetFileReader(RandomAccessFile randomAccessFile)
-            : this(randomAccessFile, ReaderProperties.GetDefaultReaderProperties())
+            : this(randomAccessFile, null)
         {
         }
 
-        public ParquetFileReader(string path, ReaderProperties readerProperties)
+        public ParquetFileReader(string path, ReaderProperties? readerProperties)
```
adamreeve, line 19: Nullability of reference types isn't part of the binary method signature so this is ABI compatible as well as source compatible.

csharp/ColumnDescriptor.cs
```diff
@@ -69,11 +69,22 @@ public TReturn Apply<TReturn>(LogicalTypeFactory typeFactory, Type? columnLogica
             var (physicalType, logicalType) = typeFactory.GetSystemTypes(this, columnLogicalTypeOverride);
             var elementType = logicalType;
 
-            for (var node = SchemaNode; node != null; node = node.Parent)
+            var node = SchemaNode;
+            while (node != null)
             {
-                if (node.LogicalType.Type == LogicalTypeEnum.List)
+                try
```
marcin-krystianc, line 75: I don't think that `try/finally` improves anything here. In my opinion it just adds complexity to the code without any real benefit.

csharp/Schema/GroupNode.cs
```diff
@@ -37,11 +37,24 @@ public int FieldIndex(Node node)
...
-                LogicalType is NoneLogicalType ? null : LogicalType);
+            using var logicalType = LogicalType;
+            var fields = Fields;
+            var clonedFields = fields.Select(f => f.DeepClone()).ToArray();
+            try
+            {
+                return new GroupNode(
+                    Name,
+                    Repetition,
+                    fields,
```
marcin-krystianc, line 48:
```suggestion
                    clonedFields,
```

csharp/Schema/GroupNode.cs
```diff
@@ -37,11 +37,24 @@ public int FieldIndex(Node node)
...
+                    logicalType is NoneLogicalType ? null : logicalType);
+            }
+            finally
+            {
+                foreach (var field in fields.Concat(clonedFields))
```
marcin-krystianc, line 53: Is it correct to dispose `clonedFields` here ?

csharp.test/TestByteArrayReaderCache.cs
```diff
@@ -29,7 +29,8 @@ public static void TestReadingDuplicateStrings([Values(true, false)] bool enable
             // Write a file that contains a lot of duplicate strings.
             using (var output = new BufferOutputStream(buffer))
             {
-                using var fileWriter = new ParquetFileWriter(output, columns, CreateWriterProperties(enableDictionary));
+                using var writerProperties = CreateWriterProperties(enableDictionary);
```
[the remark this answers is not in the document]
adamreeve: Yeah this pattern is used a lot throughout the tests when writing and then needing to dispose the writing related objects before reading, so I wouldn't want to change that here. It also doesn't bother me too much, I think it's fairly obvious in both cases when the different objects will be disposed.

csharp/ColumnDescriptor.cs
adamreeve: So you mean just remove the try/finally and keep the rest of the code as-is, so if there's an exception we let the finalizers run to free the handles? That seems reasonable to me, especially as this code is highly unlikely to throw.

csharp/Schema/GroupNode.cs
adamreeve: A whoops, good catch thanks.

csharp/Schema/GroupNode.cs
adamreeve: Yes, once we've created the GroupNode, it no longer references the original C# side fields but internally in the C++ GroupNode it has shared pointers that point to the fields. Disposing of the field just deletes the shared pointer to the C++ field, but the field itself won't be deleted due to the group node still holding a shared pointer to it.

csharp/ColumnDescriptor.cs
adamreeve: Do you think I should similarly remove `try/finally` elsewhere, eg. in the tests?

csharp/Schema/GroupNode.cs
marcin-krystianc: 👍

csharp/ColumnDescriptor.cs
adamreeve: Right, good point, I've changed this now.

marcin-krystianc (approved):

PR #295, opened 2022-08-08
Read whole leaf-level arrays at once when possible, rather than growing lists

adamreeve:
This PR changes how array values are read to avoid growing lists dynamically, and instead create arrays directly from spans of the buffered logical values.

This is based on the work done by @philjdf in https://github.com/philjdf/ParquetSharp/pull/4 and https://github.com/philjdf/ParquetSharp/pull/5. The new benchmark has mostly been copied as-is, but the array reading changes have been adapted to work with the latest master code.

On my machine (AMD Ryzen 9 5900 X with 64 GB RAM running Fedora Linux 36 with dotnet SDK 6.0.107), I get the following benchmark results:
| Test case | Time taken (s) | Time taken with server GC (s) |
| --- | --- | --- |
| Parquet.Net | 10.8 | 10.1 | 
| ParquetSharp 7.0 | 20.4 | 20.9 |
| ParquetSharp 8.0 beta | 12.5 | 10.9 |
| ParquetSharp with this PR | 4.5 | 2.6 |

The speed up from ParquetSharp 7.0 to 8.0 looks to be due to changes in Arrow rather than any of the changes we've made in ParquetSharp. I've also included benchmark results when enabling the dotnet server GC as this can improve performance with ParquetSharp significantly.

### DIFF
--- a/csharp/LogicalColumnReader.cs
+++ b/csharp/LogicalColumnReader.cs
@@ -286,36 +286,47 @@ namespace ParquetSharp
 
         private static Array ReadArrayLeafLevel(Node node, BufferedReader<TLogical, TPhysical> valueReader, short repetitionLevel, short definitionLevel)
         {
-            var values = new List<TLogical>();
-            var firstValue = true;
+            var valueChunks = new List<TLogical[]>();
             var innerNodeIsOptional = node.Repetition == Repetition.Optional;
             definitionLevel += (short) (innerNodeIsOptional ? 1 : 0);
 
+            var atArrayStart = true;
             while (!valueReader.IsEofDefinition)
             {
-                var defn = valueReader.GetCurrentDefinition();
-
-                if (!firstValue && defn.RepLevel < repetitionLevel)
+                var reachedArrayEnd =
+                    valueReader.ReadValuesAtRepetitionLevel(repetitionLevel, definitionLevel, atArrayStart,
+                        out var valuesSpan);
+                if (reachedArrayEnd && atArrayStart)
+                {
+                    return valuesSpan.ToArray();
+                }
+                atArrayStart = false;
+                valueChunks.Add(valuesSpan.ToArray());
+                if (reachedArrayEnd)
                 {
                     break;
                 }
-
-                if (defn.DefLevel == definitionLevel || innerNodeIsOptional)
-                {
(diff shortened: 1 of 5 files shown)


philjdf (approved):

philjdf:
Nice speed improvement, thanks Adam!

csharp/LogicalColumnReader.cs
```diff
@@ -286,36 +286,47 @@ public override int ReadBatch(Span<TElement> destination)
...
+                if (reachedArrayEnd && atArrayStart)
                 {
-                    // Note that when the inner node is optional,
-                    // the converter will handle definition levels and create null values.
-                    values.Add(valueReader.ReadValue());
+                    return valuesSpan.ToArray();
                 }
-                else
+                atArrayStart = false;
+                valueChunks.Add(valuesSpan.ToArray());
```
marcin-krystianc, line 304: Aren't we copying data unnecessarily here? Would it be possible to further improve the performance by using different data type for `valueChunks`, e.g.: `List<(TLogical[] logicalValues, ind index, int size)>`  and avoid creating temporary arrays entirely?

marcin-krystianc (commented):
Nice improvement. See my comment, maybe it is possible to make it even better?

csharp/LogicalColumnReader.cs
adamreeve: No, we need to convert the span to an array before the next call to `ReadValuesAtRepetitionLevel`. The span is pointing into a buffer that will be reused for the next operation, which will overwrite the previously read values.

As long as the array sizes are much smaller than the buffer size (which defaults to 4096 but is configurable), most of the time we should hit the above `reachedArrayEnd && atArrayStart` condition so only do the one copy.

marcin-krystianc:
I've experimented with some further optimisations, but I didn't see an actual improvement.
The most promising approach is https://github.com/G-Research/ParquetSharp/compare/master...marcin-krystianc:ParquetSharp:marcink-20220810-arraypool?expand=1&w=1 where I used ArrayPool to reduce allocations for temporary arrays. Unfortunately in benchmark it shows no actual improvement, I think it is mainly because temporary arrays are not used that often as most of the times there is a single chunk being created and returned.

PR #301, opened 2022-08-29
Support round-tripping nested values

adamreeve:
This adds supports for writing values with a nested structure, ie. where there is a group node in the Parquet file schema that doesn't represent a list or map logical type. ~~This also introduces a breaking change when reading nested data, as now we will~~ When data is nested, there's now the option to read values wrapped in a new `Nested` type that is required to interpret the data correctly when there are multiple levels of nullability. Previously the actual level at which a value was null was ambiguous.

There's a lot more context in the comments on #221

This is based on @philjdf's work in https://github.com/philjdf/ParquetSharp/pull/4 but changed in a few ways:
* Generic classes are used to represent readers and writers for different element types rather than delegates that work with non-generic `Array` values, to hopefully make things more readable and debuggable, although this implementation is a fair bit more verbose
* Works with the performance improvement when reading nested array values from #295. Note that arrays of nested values will still use a list growing approach and need to read one value at a time.
* Read and write whole batches of required nested values in one operation
* When writing optional nested values, write runs of all null or all non-null values in one operation

There's also a breaking change in the `LogicalColumnStream` API where the protected `SchemaNodesPath` property has been replaced with `GetSchemaNodesPath`, but I don't expect many (probably any) users would be affected by this as I'm not aware of any use-cases for subclassing `LogicalColumnStream`.

Benchmark results vs the current master on my machine don't show any significant performance difference:
| Benchmark | Baseline | After |
| --- | ---: | ---: |
| DecimalRead | 82.18 +- 4.32 ms | 80.69 +- 1.323 ms |
| DecimalWrite | 112.3 +- 0.73 ms | 117.2 +- 0.68 ms |
| FloatTimeSeriesRead | 171.0 +- 3.29 ms | 161.5 +- 3.12 ms |
| FloatTimeSeriesWrite | 63.73 +- 0.503 ms | 65.62 +- 0.611 ms |
| FloatTimeSeriesWrite - chunked | 291.04 +- 4.219 ms | 291.9 +- 3.711 ms |
| FloatArrayTimeSeriesRead | 4.042 +- 0.0863 s | 3.95 +- 0.0581 s |


adamreeve:
After some discussion with @marcin-krystianc, we thought it would be nice if we could introduce this change in a backwards compatible way so that users could continue to read nested data without needing the `Nested` wrapper type, so I've just pushed a change that does this. The `Nested` type still needs to be used if writing nested data, which wasn't possible before, but code that previously read nested data should now continue to work unchanged.

Users can opt in to using the `Nested` type by using it in the `TElement` parameter of the `LogicalColumnReader<TElement>` method if they have a need to accurately reproduce the nested structure of the data, or if using the non-generic `LogicalColumnReader` method there's a new override with a `useNesting` parameter.

marcin-krystianc (dismissed):
Well done, I'm glad it was possible to maintain backwards compatibility for reading.

jgiannuzzi:
Hi @philjdf and @adeboyed, we would love to have you review this PR before we merge it — let us know whether that's possible for you!

docs/Nested.md
```diff
@@ -0,0 +1,244 @@
...
+
+ParquetSharp's API works with leaf-level columns,
+so in dotnet the map data cannot be represented as a column of
+`Dictionary` objects, and instead is read as two separate columns.
+The first contains arrays of the map keys,
+and the second contains arrays of the map values,
+and the arrays corresponding to the same row must have the same length.
+
+The following example shows how dotnet dictionary data might be written
+and then read from Parquet::
```
philjdf, line 192: `::`?

philjdf (approved):
Looks very good to me, particularly like the LogicalBatchReader.cs and LogicalBatchWriter.cs files, and that it also fixes required arrays etc.

docs/Nested.md
adamreeve: Whoops, must have got confused with rst syntax, fixed now

marcin-krystianc (approved):

PR #302, opened 2022-09-07
Allow NullLogicalType given a PhysicalType

jgiannuzzi:
_This PR is on behalf of @msumner91._

The goal is to fix the following exception:
```
System.ArgumentOutOfRangeException: unsupported logical type Null with physical type Int32
Parameter name: logicalType
at ParquetSharp.LogicalTypeFactory.GetSystemTypes(ColumnDescriptor descriptor)
at ParquetSharp.LogicalTypeFactory.GetSystemTypes(ColumnDescriptor descriptor, Type columnLogicalTypeOverride)
at ParquetSharp.ColumnDescriptor.GetSystemTypes(LogicalTypeFactory typeFactory, Type columnLogicalTypeOverride)
```

### DIFF
--- a/csharp.test/TestCaseGenericAttribute.cs
+++ b/csharp.test/TestCaseGenericAttribute.cs
@@ -1,0 +1,36 @@
+using System;
+using System.Collections.Generic;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+using NUnit.Framework.Internal.Builders;
+
+namespace ParquetSharp.Test
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+    internal sealed class TestCaseGenericAttribute : TestCaseAttribute, ITestBuilder
+    {
+        public TestCaseGenericAttribute(params object[] arguments)
+            : base(arguments)
+        {
+        }
+
+        public Type[] TypeArguments { get; set; } = null!;
+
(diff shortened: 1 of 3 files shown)


jgiannuzzi:
@msumner91 gave the following code to create a Parquet file to reproduce the issue:
```python
import pandas as pd
dd = pd.DataFrame([[None]], columns=["matchingFilters"])
dd.to_parquet("bad.parquet")
```

adamreeve:
I've added a new unit test to cover this, and also added a check that this is only used for optional columns, as my understanding is that the null logical type should only be used when all values in a column are null.

adamreeve (dismissed):

marcin-krystianc (approved):

jgiannuzzi:
Thanks for adding the unit test @adamreeve!
I've parametrised it with @mharward-gr's precious help. Could you please review again @marcin-krystianc?

marcin-krystianc (approved):

adamreeve:
> Thanks for adding the unit test @adamreeve! I've parametrised it with @mharward-gr's precious help.

Nice! I wasn't aware you could have generic test cases like that

adamreeve:
> FYI, I've tried writing null and non-null values and it worked just fine. Not sure, whether it is a problem or not.

Yeah, we don't validate that only null values are written in ParquetSharp and it looks like the Arrow C++ library doesn't do this either, but the intention is that this logical type is used when the actual type can't be inferred as all values are null (like in the example Python code above).

PR #303, opened 2022-09-09
Bump version to 8.0.0-beta2

jgiannuzzi:


m4rs-mt (approved):
As discussed offline, this should be good to go.

Issue #114, opened 2020-03-06
Long filename support


philjdf:
ParquetFileReader when used with a long string path fails:
```
ParquetSharp.ParquetException : class parquet::ParquetStatusException (message: 'IOError: Failed to open local file '<very long filename>'. Detail: [Windows error 3] The system cannot find the path specified.
')
   at ParquetSharp.ExceptionInfo.Check(IntPtr exceptionInfo)
   at ParquetSharp.ParquetFileReader..ctor(String path, ReaderProperties readerProperties)
   ...
```

Looks like ParquetSharp eventually calls https://github.com/apache/arrow/blob/88e3267ad09ca62643b7fe5ffd98eb03b29728fc/cpp/src/arrow/util/io_util.cc#L838 which is calling `_wsopen_s` on Windows, so https://docs.microsoft.com/en-us/cpp/c-runtime-library/path-field-limits?view=vs-2019 appears relevant, so it is just a case of needing to prepend the filename with `\\?\`?

GPSnoopy:
@philjdf I believe we've seen this one before. ParquetSharp is irrelevant. The application itself needs to support long paths. 

IIRC .NET Core 3 will work out of the box. .NET Framework 4.6 or so needs playing around with the application manifest. 

Google and small trial applications are you friends.

GPSnoopy:
I may have spoken too soon. It indeed does not seem to work under .NET Core 3.1 (at least not straight out of the box).

GPSnoopy:
It looks like the proper Win32 API needs to be used to expose the long path support:
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew

In particular: 

> Tip  Starting with Windows 10, version 1607, for the unicode version of this function (CreateFileW), you can opt-in to remove the MAX_PATH limitation without prepending "\\?\". See the "Maximum Path Length Limitation" section of Naming Files, Paths, and Namespaces for details.

GPSnoopy:
See more details at:
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#enable-long-paths-in-windows-10-version-1607-and-later

Worth a try in Arrow: use Win32 API to open the file, but use `_get_osfhandle` to return a file descriptor.

tpboudreau:
Sorry for taking so long to get back to this.  

[...]

PR #311, opened 2022-10-25
Convert paths to extend length format on Windows and add documentation on long path support

adamreeve:
Although Arrow was updated to use long-path aware win32 APIs (#114), there are additional configuration requirements needed to handle long paths, so I've added documentation on this.

In addition, paths need to be provided in extended-length format, even in dotnet core, so I've added code to do this conversion automatically for absolute paths.


marcin-krystianc (approved):
👍

PR #320, opened 2022-11-28
Allow creating a ParquetFileReader/Writer directly from a .NET stream

adamreeve:
This PR adds some convenience constructors that allow reading or writing a .NET stream directly without having to create a `ManagedRandomAccessFile` or `ManagedOutputStream` explicitly.


marcin-krystianc (approved):
LGTM

PR #321, opened 2022-12-04
Upgrade to Arrow 10.0.1

adamreeve:
This upgrades the version of Arrow used to 10.0.1 and bumps the ParquetSharp version to 10.0.0-beta1.

The main user-visible change with this is that the default version of the Parquet format to write is now 2.4. The Arrow Jira issue that explains the motivation for this is https://issues.apache.org/jira/browse/ARROW-12203.

I believe the points about supported time and integer types are not relevant to us as these only affect type mappings when converting from Arrow data, and we already used these new types when writing 1.0 versioned files with the corresponding .NET types. One change that could cause issues is that dictionary encoded data now uses the `RleDictionary` encoding by default rather than `PlainDictionary`, which could cause compatibility issues with old versions of some libraries. Eg. Parquet.Net only supports `RleDictionary` since 4.0.2, which also dropped .NET framework support. I believe this isn't actually a different encoding implementation to `PlainDictionary` but just a naming tidy up.

It's worth pointing out that we could upgrade to Arrow 10.0.1 but override the default version to write to keep it as 1.0 if we wanted, but it seems sensible to follow the lead of the Arrow project here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ In the 5.0.X versions, reading nested structures was introduced. However, nestin
 
 Building ParquetSharp for Windows requires the following dependencies:
 - Visual Studio 2022 (17.0 or higher)
-- Apache Arrow (8.0.0)
+- Apache Arrow (10.0.1)
 
 For building Arrow (including Parquet) and its dependencies, we recommend using Microsoft's [vcpkg](https://github.com/Microsoft/vcpkg).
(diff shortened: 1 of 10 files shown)


marcin-krystianc (approved):
> It's worth pointing out that we could upgrade to Arrow 10.0.1 but override the default version to write to keep it as 1.0 if we wanted, but it seems sensible to follow the lead of the Arrow project here.

I agree we should align with the Arrow project as much as possible.

PR #330, opened 2023-01-25
Use ISO 8601 date format for Date.ToString

adamreeve:


marcin-krystianc (approved):

csharp/Date.cs
```diff
@@ -65,7 +66,7 @@ public int CompareTo(Date other)
 
         public override string ToString()
         {
-            return DateTime.ToString();
+            return DateTime.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture);
```
marcin-krystianc, line 69: This change is somewhat backward incompatible, but I think that in this case the benefits outweigh to costs.

adamreeve: Yeah we did something similar in https://github.com/G-Research/ParquetSharp/pull/262 and decided that change should be part of a major release due to changing behaviour, so I think the same reasoning applies here. I should have thought to do the `Date` type at the same time then.

PR #348, opened 2023-04-20
Defer checking decimal precision and type length until creating converter

### LINKED ISSUE #347: Allow logical type override for decimal when logical system type (c# type) is not decimal
Currently when we create ParquetSharp column for decimal type, it only allows decimal logical type override for Decimal128. However, we are extending the type factory and converter factory to support other precision of decimal and due to this check, we cannot define column with our own implemented BigDecimal based on decimal logical type override. Instead, we have to create the schema node which is kind of hack as currently the offical extension support is to extend type factory and converter factory. Suggest to add one more check in the condition as below which only checks for c# decimal:
```
if (logicalTypeOverride is DecimalLogicalType decimalLogicalType && (logicalSystemType == typeof(decimal) || logicalSystemType == typeof(decimal?))
{
    // For the moment we only support serializing decimal to Decimal128.
    // This reflects the C# decimal structure with 28-29 digits precision.
    // Will implement 32-bits, 64-bits and other precision later.
    if (decimalLogicalType.Precision != 29)
    {
        throw new NotSupportedException("only 29 digits of precision is currently supported for decimal type");
    }

    if (length != sizeof(Decimal128))
    {
        throw new NotSupportedException("only 16 bytes of length is currently supported for decimal type ");
    }
}
```

adamreeve:
This allows use of custom converters that support different precision and type lengths.

Fixes #347

### DIFF
--- a/csharp.test/TestDecimal128.cs
+++ b/csharp.test/TestDecimal128.cs
@@ -98,5 +98,35 @@ namespace ParquetSharp.Test
             var read = (decimal[]) rowGroupReader.ReadColumn(fileReader.Schema.GetDataFields()[0]).Data;
             Assert.AreEqual(values, read);
         }
+
+        [Test]
+        public static void TestThrowsWithUnsupportedPrecision()
+        {
+            using var decimalType = LogicalType.Decimal(precision: 28, scale: 3);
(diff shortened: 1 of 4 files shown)


marcin-krystianc (approved):
👍

PR #349, opened 2023-04-20
Add WriteRowSpan method to ParquetRowWriter

adamreeve:
This method improves performance by reducing the number of allocations needed. In the new benchmark added I see a 25% speed up.

### DIFF
--- a/csharp/RowOriented/ParquetRowWriter.cs
+++ b/csharp/RowOriented/ParquetRowWriter.cs
@@ -95,6 +95,19 @@ namespace ParquetSharp.RowOriented
             }
         }
 
+        public void WriteRowSpan(ReadOnlySpan<TTuple> rows)
+        {
+            if (_pos + rows.Length > _rows.Length)
+            {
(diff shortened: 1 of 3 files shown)


csharp/RowOriented/ParquetRowWriter.cs
```diff
@@ -95,6 +95,19 @@ public void WriteRows(IEnumerable<TTuple> rows)
             }
         }
 
+        public void WriteRowSpan(ReadOnlySpan<TTuple> rows)
```
adamreeve, line 98: I initially tried to add this as an override of the existing `WriteRows` method, but this is not source backwards-compatible, as any existing calls to `WriteRows` using an array would be ambiguous between the overload that takes an `IEnumerable` and this new method taking a `ReadOnlySpan`. Places where `WriteRows(array)` is used would need to be changed to `WriteRows(array.AsSpan())` to disambiguate and use the new method. So using a new method name seems like a better approach.

marcin-krystianc (approved):
👍

PR #354, opened 2023-06-09
Support reading and writing data with the Arrow format

adamreeve:
This PR adds support for reading Parquet files as Arrow data, and writing Arrow data to Parquet.

This uses the Arrow reading and writing functionality from the Arrow Parquet C++ library and passes Arrow data between C++ and .NET using the Arrow C Data Interface, without needing to copy it.

I've also added new benchmarks that use Arrow data and fixed the benchmarks so that they run when `Check.Enabled` is set to true.


csharp/Arrow/ArrowTimeUnit.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+
+namespace ParquetSharp.Arrow
+{
+    /// <summary>
+    /// Arrow TimeUnit enum, with values corresponding to those used in the C++ library.
+    /// These do match the C# Arrow library enum values but there's no guarantee that will always be true.
```
marcin-krystianc, line 7: Do I understand this correctly? If we had a guarantee that values in the C# Arrow library and C++ library match, we wouldn't have to implement the `ArrowTimeUnit`? Maybe there is no guarantee, but it would be madness if C++ and C# libraries defined different enum values, so I find this an overkill. (But I might be wrong, so feel free to ignore my opinion).

marcin-krystianc (approved):

csharp/Arrow/ArrowTimeUnit.cs
adamreeve: Yeah, you're right, I was probably overthinking this. I will simplify this and remove the extra enum. I think it might be sensible to have a unit test to verify the enum values match though

csharp/Arrow/FileReader.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// </summary>
+        /// <param name="path">Path to the Parquet file</param>
+        /// <param name="properties">Parquet reader properties</param>
+        /// <param name="arrowProperties">Arrow specific reader properties</param>
+        public FileReader(
+            string path,
+            ReaderProperties? properties = null,
+            ArrowReaderProperties? arrowProperties = null)
+        {
+            using var defaultProperties = properties == null ? ReaderProperties.GetDefaultReaderProperties() : null;
```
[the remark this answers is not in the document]
adamreeve: These aren't really equivalent, we don't want `defaultProperties` to be set to `properties` when it's non-null, as we don't want to dispose of the user-provided properties.

marcin-krystianc (approved):

PR #363, opened 2023-08-09
fix ci warnings

ljubon:
- [ ] remove `.NET 5.0`
- [ ] add `.NET 7.0`
- [ ] no ci warnings

CI workflow without worning - https://github.com/ljubon/ParquetSharp/actions/runs/5801944212/job/15727971495

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
   check-format:
     if: github.event_name == 'schedule' || github.event_name == 'push' || github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id
     name: Check format
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -49,11 +49,14 @@ jobs:
     if: github.event_name == 'schedule' || github.event_name == 'push' || github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id
     strategy:
       matrix:
-        os: [ubuntu-20.04, macos-11, windows-2022]
+        os: [ubuntu-22.04, macos-11, windows-2022]
```
adamreeve, line 52: I think we should stick with ubuntu-20.04 here for now so that the Arm native library is compatible with ubuntu 20.04, meaning we can remove the extra include below for arm64 on ubuntu-20.04.

It doesn't make sense to have both ubuntu-22.04 arm64 and ubuntu-20.04 arm64, as these will both have the same value for `$triplet` so be trying to create the same artifact name.

.github/workflows/ci.yml
```diff
@@ -127,21 +127,21 @@ jobs:
 
     # Install arm64 cross-compilation toolchain if required
     - name: Install arm64 cross-compilation toolchain
-      if: runner.os == 'Linux' && matrix.arch == 'arm64'
+      if: runner.os == 'Linux' && matrix.os == 'ubuntu-20.04'
```
adamreeve, line 130: This  if condition should be kept as it was, we don't need to install these on the x64 build.

Also, why the change to the packages installed below? It looks like `g++-aarch64-linux-gnu` depends on `gcc-aarch64-linux-gnu` so we shouldn't need to depend on that explicitly. Given the build is working, maybe we don't need `pkg-config-aarch64-linux-gnu`, but I'd be more comfortable leaving that is.

adamreeve (approved):

PR #365, opened 2023-08-10
Finalize net7.0 migration

jgiannuzzi:
#363 started the migration to `net7.0` but stopped short of running actual tests and building with the `net7.0` SDK — this PR aims at fixing that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,15 +244,15 @@ jobs:
       uses: actions/setup-dotnet@v3
       with:
         dotnet-version: 3.1.x
-    - name: Setup .NET SDK v7.0.x
-      if: matrix.dotnet == 'net7.0'
-      uses: actions/setup-dotnet@v3
-      with:
-        dotnet-version: 7.0.x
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -244,15 +244,15 @@ jobs:
...
-      if: matrix.dotnet == 'net7.0'
-      uses: actions/setup-dotnet@v3
-      with:
-        dotnet-version: 7.0.x
     - name: Setup .NET SDK v6.0.x
+      if: matrix.dotnet == 'net6.0'
       uses: actions/setup-dotnet@v3
       with:
         dotnet-version: 6.0.x
+    - name: Setup .NET SDK v7.0.x
```
adamreeve, line 252: Does this need an `if: matrix.dotnet == 'net7.0'`?

jgiannuzzi: No because we want to install this SDK in all cases, as it's required to build any target now that we have updated our `global.json`

adamreeve (approved):

marcin-krystianc (approved):

PR #368, opened 2023-08-30
Upgrade to Arrow 13.0.0

adamreeve:
This upgrades the version of the Arrow C++ library to 13.0.0.

The notable changes in this version are:
* Writing the page index can now be controlled per column
* The Parquet version written now defaults to 2.6. This version includes the nanosecond timestamp precision but we were already using this in ParquetSharp so this change doesn't seem to affect us besides the metadata being different
* Changes to how the column encodings are computed affected the order these are returned in, but the order shouldn't have been relied on

Comparing benchmark results between the current master and this branch, the timings are mostly the same, with some of the "chunked" writing tests slowing down a little (5 to 10%) indicating there might be a bit more per-row-group overhead. There is quite a large slowdown in the nested write test though, going from 47 ms to 85 ms per iteration on my machine. I looked into this and have made a PR to fix this in Arrow (https://github.com/apache/arrow/pull/37454), so hopefully the performance will be better again for the next version. I think this regression probably isn't bad enough to block upgrading to 13.0.0 though as it looks like it should only affect writing nested data, where we make lots of calls to WriteBatch with small batches of data. (And possibly this could be refactored to avoid needing to make so many WriteBatch calls if we buffered these first?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ As only 64-bit runtimes are available, ParquetSharp cannot be referenced by a 32
 
 Building ParquetSharp for Windows requires the following dependencies:
 - Visual Studio 2022 (17.0 or higher)
-- Apache Arrow (12.0.1)
+- Apache Arrow (13.0.0)
 
 For building Arrow (including Parquet) and its dependencies, we recommend using Microsoft's [vcpkg](https://github.com/Microsoft/vcpkg).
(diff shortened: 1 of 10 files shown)


marcin-krystianc (approved):
Looks good 👍

PR #371, opened 2023-09-09
Split up LogicalBatchReader/Writer into separate files

philjdf:
Just a small PR so it's easier to navigate around that part of the code (with a view to subsequently having a go at adding a `Skip` method to `LogicalColumnReader`).


adamreeve (approved):
:+1: thanks Phil

PR #373, opened 2023-09-14
Document workarounds for using the dev container with podman and SELinux

adamreeve:
This documents the workarounds I had to use to get the dev container working on Fedora 38 with SELinux and podman. I've put these in a collapsible section as they won't be relevant for most users, but it seems worth having them documented here in case anybody else has similar issues.


marcin-krystianc (approved):
Looks good 👍

jgiannuzzi (approved):
Excellent, thanks for documenting this Adam!

PR #378, opened 2023-10-05
TestWriterDoubleDispose: remove outdated comment

mfkl:
Just a minor cleanup, getting familiar with the project :-)

### DIFF
--- a/csharp.test/TestRowOrientedParquetFile.cs
+++ b/csharp.test/TestRowOrientedParquetFile.cs
@@ -113,9 +113,6 @@ namespace ParquetSharp.Test
         [Test]
         public static void TestWriterDoubleDispose()
         {
-            // ParquetRowWriter is not double-Dispose safe (Issue 64)
-            // https://github.com/G-Research/ParquetSharp/issues/64
-
             using var buffer = new ResizableBuffer();
             using var outputStream = new BufferOutputStream(buffer);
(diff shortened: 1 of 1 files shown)


csharp.test/TestRowOrientedParquetFile.cs
```diff
@@ -113,14 +113,12 @@ public static void TestEmptyRowGroup([Values(false, true)] bool closeBeforeDispo
         [Test]
         public static void TestWriterDoubleDispose()
         {
-            // ParquetRowWriter is not double-Dispose safe (Issue 64)
-            // https://github.com/G-Research/ParquetSharp/issues/64
-
             using var buffer = new ResizableBuffer();
             using var outputStream = new BufferOutputStream(buffer);
             using var writer = ParquetFile.CreateRowWriter<(int, double, DateTime)>(outputStream);
```
adamreeve, line 118: I think this was already testing the double dispose due to the `using var` statement here, although the comment is quite misleading, it should probably say something like "Test that ParquetRowWriter is double-Dispose safe"

adamreeve: See the commit where this test was added with the fix here: https://github.com/G-Research/ParquetSharp/commit/ce713348566d62c8959ba772531bfe84405e9424

mfkl: Yep of course it was 🤦‍♂️

adamreeve (approved):

PR #384, opened 2023-10-16
Write vcpkg arrow port ABI info in CI

adamreeve:
This should help us understand why vcpkg sometimes fails to use cached previous builds

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,6 +177,8 @@ jobs:
       with:
         name: ${{ steps.vcpkg-info.outputs.triplet }}-vcpkg-arrow-logs
         path: ${{ steps.vcpkg-info.outputs.root }}/buildtrees/arrow/*.log
+    - name: Dump vcpkg arrow abi info
+      run: cat ./build/${{ steps.vcpkg-info.outputs.triplet }}-release/vcpkg_installed/${{ steps.vcpkg-info.outputs.triplet }}/share/arrow/vcpkg_abi_info.txt
     - name: Build .NET benchmarks & unit tests
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -177,6 +177,12 @@ jobs:
       with:
         name: ${{ steps.vcpkg-info.outputs.triplet }}-vcpkg-arrow-logs
         path: ${{ steps.vcpkg-info.outputs.root }}/buildtrees/arrow/*.log
+    - name: Dump vcpkg arrow abi info (Unix)
+      if: runner.os == 'Linux' || runner.os == 'macOS'
+      run: cat ./build/${{ steps.vcpkg-info.outputs.triplet }}-release/vcpkg_installed/${{ steps.vcpkg-info.outputs.triplet }}/share/arrow/vcpkg_abi_info.txt
+    - name: Dump vcpkg arrow abi info (Windows)
+      if: runner.os == 'Windows'
+      run: Get-Content ./build/${{ steps.vcpkg-info.outputs.triplet }}/vcpkg_installed/${{ steps.vcpkg-info.outputs.triplet }}/share/arrow/vcpkg_abi_info.txt
```
marcin-krystianc, line 185: nit: I think it is possible to use powershell or unix tools on all OSes, so there is no need to do it conditionally. But doing it conditionally is fine too.

marcin-krystianc (approved):

.github/workflows/ci.yml
adamreeve: So I originally just had one job for all OSs, as powershell recognizes  `cat` as an alias for `Get-Content`, but the Windows path is slightly different and doesn't have the extra "-release". I could probably figure out how to add that conditionally, but it seemed easier just to have separate jobs.

Issue #239, opened 2021-12-14
How can I use ParquetSharp from PowerShell?


dburtsev:
Hello
I would like to update some of my ETL scripts from csv to parquet.
1. How can I install ParquetSharp on the server without VS?
2. How can I reference ParquetSharp from the script (Import-Module / Add-Type -Assembly / [System.Reflection.Assembly]::LoadWithPartialName) ?
Thanks,
Dmitriy

adamreeve:
Hi Dmitriy

ParquetSharp is available via NuGet so you can install the NuGet command line application and use that to install ParquetSharp, see the [NuGet command line docs here](https://docs.microsoft.com/en-us/nuget/reference/nuget-exe-cli-reference).

I tested using ParquetSharp from PowerShell and it is possible, one extra step that's needed is setting your Path environment variable so the native dll can be loaded. For example:
```powershell
# Install ParquetSharp into the current directory
nuget.exe install ParquetSharp

# Add native dll directory to Path
$nativeDllDir = Join-Path  (Get-Location) "ParquetSharp.4.0.0\runtimes\win-x64\native"
$env:Path += ";$nativeDllDir"

# Set current directory for .Net to match PowerShell
[Environment]::CurrentDirectory = (Get-Location).Path

# Load ParquetSharp
Add-Type -Path .\ParquetSharp.4.0.0\lib\net471\ParquetSharp.dll

# Create a ParquetFileReader
$reader = New-Object -TypeName ParquetSharp.ParquetFileReader -ArgumentList "./data.parquet"
```

I imagine trying to use the ParquetSharp API from PowerShell could get quite awkward though, you'd probably be better off writing an application in C# or F# that you could call from PowerShell.

jgiannuzzi:
Thanks for chiming in, @gitfvb! Would you agree for us to link to your examples in our README?

gitfvb:
Hi @jgiannuzzi , sure. It is an honour for me. Feel free to link me or copy&paste the code you need 😊

dburtsev:
Thank you adamreeve
Can I run _nuget.exe install ParquetSharp_ on my Win10 PC and copy ParquetSharp.dll to the windows server?

>you'd probably be better off writing an application in C#

I am the only one in my team who knows C#. This is why I have to use PowerShell.

dburtsev:
1. I create a directory C:\Lib
2. Run nuget.exe install ParquetSharp.
[...]

PR #386, opened 2023-10-21
Enhance PowerShell documentation

jescalada:
This is another PR for enhancing the available documentation. I felt that the PowerShell guide could use some improvements to reduce friction for new users, so I simplified the installation and basic usage steps. This guide still assumes that the user is familiar with PowerShell scripting.  

Feel free to make any changes as required.

## Changes
- Clarified installation steps using nuget (for Windows users specifically)
- Added extra requirements for scripts to execute (editing versions, selecting architecture)
  - Required for `Copy-Item` snippets to work correctly
  - Added reminder for user to change the version/architecture
- Added basic powershell usage: loading `ParquetSharp.dll` and getting a reference to a `ParquetFileReader`
  - Lists the previous example script if user wants more advanced functionality

### DIFF
--- a/docs/PowerShell.md
+++ b/docs/PowerShell.md
@@ -1,8 +1,17 @@
 # ParquetSharp in PowerShell
 
+### Installation
+
 It's possible to use ParquetSharp from PowerShell.
-You can install ParquetSharp with the [NuGet command line interface](https://docs.microsoft.com/en-us/nuget/reference/nuget-exe-cli-reference),
-then use `Add-Type` to load `ParquetSharp.dll`.
+You can install ParquetSharp with the [NuGet command line interface](https://docs.microsoft.com/en-us/nuget/reference/nuget-exe-cli-reference).  
+First, make sure `nuget.exe` is in your `PATH`, then run the following to install the latest version of ParquetSharp into the lib directory:
+
+```powershell
+  nuget install ParquetSharp -OutputDirectory lib
+```
+
+### Usage
+Use `Add-Type` to load `ParquetSharp.dll`.
 However, you must ensure that the appropriate `ParquetSharpNative.dll` for your architecture and OS can be loaded as required,
 either by putting it somewhere in your `PATH` or in the same directory as `ParquetSharp.dll`.
 For examples of how to use ParquetSharp from PowerShell,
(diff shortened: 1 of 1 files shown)


docs/PowerShell.md
```diff
@@ -1,10 +1,56 @@
...
 It's possible to use ParquetSharp from PowerShell.
-You can install ParquetSharp with the [NuGet command line interface](https://docs.microsoft.com/en-us/nuget/reference/nuget-exe-cli-reference),
-then use `Add-Type` to load `ParquetSharp.dll`.
-However, you must ensure that the appropriate `ParquetSharpNative.dll` for your architecture and OS can be loaded as required,
-either by putting it somewhere in your `PATH` or in the same directory as `ParquetSharp.dll`.
-For examples of how to use ParquetSharp from PowerShell,
-see [these scripts from Apteco](https://github.com/Apteco/HelperScripts/tree/master/scripts/parquet).
+You can install ParquetSharp with the [NuGet command line interface](https://docs.microsoft.com/en-us/nuget/reference/nuget-exe-cli-reference).  
+
+First, make sure `nuget.exe` is in your `PATH`, or in the current directory, then run the following to install the latest version of ParquetSharp into a new directory called `lib`:
```
adamreeve, line 8: These instructions are presented as if this is the one way of using ParquetSharp from PowerShell, but there's no requirement that the dlls are in a directory named "bin" for example. I think it's worth clarifying that this is just one possible approach, and we should keep the existing text that says that the main requirement is that `ParquetSharpNative.dll` needs to be on the `PATH` or in the same directory as `ParquetSharp.dll`.

docs/PowerShell.md
```diff
@@ -1,10 +1,56 @@
...
+
+The available runtime architectures are `win-x64`, `linux-x64`, `linux-arm64`, `osx-x64`, and `osx-arm64`.
+
+### Usage
+Use `Add-Type` to load `ParquetSharp.dll`:
+
+```powershell
+$dlls = Get-ChildItem -Path ".\bin" -Filter "*.dll" | where { @("ParquetSharpNative.dll") -notcontains $_.Name }
+
+$dlls | ForEach {
```
adamreeve, line 43: Is it necessary to explicitly load each dll from the bin directory? I would expect that you can just do `Add-Type -Path bin\ParquetSharp.dll` and dlls that are dependencies will be loaded as required?

Eg. this worked for me when I originally looked into this: https://github.com/G-Research/ParquetSharp/issues/239#issuecomment-995191713

jescalada: I replaced some of the existing code with the one you detailed in the PR. I think it's a lot easier to understand. 🤔

adamreeve (approved):
:+1: thanks @jescalada

PR #388, opened 2023-10-27
add Skip at the LogicalColumnReader level

mfkl:


philjdf:
Hi @mfkl,

I think this might not work for e.g. array-valued columns, as e.g. skipping 1 row might require skipping n underlying items? Sorry if I'm missing something.

Cheers,

Phil

mfkl:
Hi Phil, I guess I misunderstood then. Will revisit the test and implementation! Thanks for the feedback.

csharp/LogicalColumnReader.cs
```diff
@@ -199,6 +201,22 @@ public override int ReadBatch(Span<TElement> destination)
             return _batchReader.ReadBatch(destination);
         }
 
+        public override long Skip(long numRowsToSkip)
+        {
+            var isArray = typeof(TElement).IsArray;
+            if (isArray)
+            {
+                var arrayReader = _batchReader as IArraySkippable;
```
mfkl, line 209: This feels unelegant, but I'm not sure how to make it better. Thoughts welcome.

csharp/LogicalBatchReader/ArrayReader.cs
```diff
@@ -140,10 +140,25 @@ public bool HasNext()
             return !_bufferedReader.IsEofDefinition;
         }
 
+        public long Skip(long numRowsToSkip)
+        {
+            for (var i = 0; i < numRowsToSkip; i++)
+            {
+                ReadInnerTypeArray();
```
adamreeve, line 147: This looks like it should work but for something like an array of integers it'll be pretty inefficient as it reads one item at a time. We should be able to do something slightly better in that case, by using similar logic to `ReadLogicalTypeArray`.

We could possibly optimize this further by supporting it at the `BufferedReader` level and avoiding calling the converter on data we are skipping, but that might get a bit too complicated.

csharp/LogicalColumnReader.cs
adamreeve: I think you might want to add a Skip method to `ILogicalBatchReader` instead and implement it for all the different reader types, as eg. this doesn't look like it will work correctly for an array within a nested reader? For simple cases like the direct and scalar reader they can implement this just by calling skip on the physical column reader.

mfkl: That was my first implementation but wasn't sure it was relevant for all other reader types. Will revert back to that.

csharp/LogicalBatchReader/ArrayReader.cs
mfkl: Ok, looking into it. Thanks!

csharp/LogicalColumnReader.cs
```diff
@@ -72,6 +72,8 @@ internal static LogicalColumnReader<TElement> Create<TElement>(ColumnReader colu
 
         public abstract TReturn Apply<TReturn>(ILogicalColumnReaderVisitor<TReturn> visitor);
 
+        public abstract long Skip(long numRowsToSkip);
```
adamreeve, line 75: I know most existing public methods don't have documentation, but it would be good to have a short documentation comment on this method that contrasts it with the physical reader skip method.

csharp/LogicalBatchReader/OptionalNestedReader.cs
```diff
@@ -52,6 +52,29 @@ public bool HasNext()
...
+            for (var i = 0; i < numRowsToSkip; ++i)
+            {
+                if (_bufferedReader.IsEofDefinition)
+                {
+                    return i;
+                }
+                var defn = _bufferedReader.GetCurrentDefinition();
+                if (defn.DefLevel >= _definitionLevel)
+                {
+                    _innerReader.ReadBatch(innerValue);
```
adamreeve, line 67: I think this can be `_innerReader.Skip(1)` instead.

csharp/LogicalBatchReader/OptionalReader.cs
```diff
@@ -53,6 +53,29 @@ public bool HasNext()
...
+            for (var i = 0; i < numRowsToSkip; ++i)
+            {
+                if (_bufferedReader.IsEofDefinition)
+                {
+                    return i;
+                }
+                var defn = _bufferedReader.GetCurrentDefinition();
+                if (defn.DefLevel >= _definitionLevel)
+                {
+                    _innerReader.ReadBatch(innerValue);
```
adamreeve, line 68:
```suggestion
                    _innerReader.Skip(1);
```

csharp/LogicalBatchReader/ArrayReader.cs
```diff
@@ -140,6 +140,40 @@ public bool HasNext()
...
+                {
+                    return i;
+                }
+
+                var defn = _bufferedReader.GetCurrentDefinition();
+                if (defn.DefLevel > _definitionLevel)
+                {
+                    if (typeof(TItem) == typeof(TLogical))
+                    {
+                        ReadLogicalTypeArray();
```
adamreeve, line 157: It will probably require a bit more duplication of logic but I think it's worth optimizing this further by adding `SkipLogicalTypeArray` and `SkipInnerTypeArray` methods that have similar logic to the read methods but avoid creating arrays and copying values.

adamreeve (commented):
This is looking good thanks @mfkl, just a few suggestions.

csharp/LogicalBatchReader/ArrayReader.cs
mfkl: Ok!

csharp/LogicalBatchReader/ArrayReader.cs
```diff
@@ -86,6 +86,24 @@ private TItem[] ReadInnerTypeArray()
...
+            var firstValue = true;
+            while (!_bufferedReader.IsEofDefinition)
+            {
+                var defn = _bufferedReader.GetCurrentDefinition();
+                if (!firstValue && defn.RepLevel <= _repetitionLevel)
+                {
+                    break;
+                }
+
+                _innerReader.ReadBatch(value);
```
adamreeve, line 102:
```suggestion
                _innerReader.Skip(1);
```

adamreeve (commented):
Just one last minor thing and then this looks good to go

adamreeve (approved):

PR #392, opened 2023-11-08
CI: enable CodeQL runs

ljubon:
Goal of this PR is to enable CodeQL runs for the ParquetSharp
https://github.com/G-Research/oss-portfolio-maturity/issues/415


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,46 @@
...
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
+
+    - run: |
+        sudo apt-get update
+        sudo apt-get --yes install flex bison
+        ./vcpkg_unix.sh
```
marcin-krystianc, line 39:
```suggestion
```

adamreeve (approved):

adamreeve:
Hi @ljubon, I realised after merging that this is quite slow due to needing to rebuild all the C++ dependencies, which is a problem if we're going to run this on all pull requests. Can you please take a look at using cached dependencies, like in the "Get cached vcpkg dependencies" step in `ci.yml`?

The "Perform CodeQL Analysis" step also seems quite slow. Would it help to move the "apt-get" commands before "Initialize CodeQL" in case these are being used in the analysis?

adamreeve:
It looks like this also needs a step similar to the "Update vcpkg builtin registry" step in our CI workflow, as the vcpkg installed on the runners might be older than we need. Eg. see the failures in #407

adamreeve:
I think it's fine to do this in a separate PR without reverting. The vcpkg update issue is probably more urgent as it causes failures now, but it isn't blocking us from merging PRs.

PR #393, opened 2023-11-15
Add support for custom conversion logic with Row-oriented API

jescalada:
### Changelog:
- Added support for custom conversion logic for Row-oriented API
  - Allow passing the `LogicalReadConverterFactory` and `LogicalTypeFactory` when invoking `ParquetFile.CreateRowReader()`
  - Allow passing the `LogicalWriteConverterFactory` and `LogicalTypeFactory` when invoking `ParquetFile.CreateRowWriter()`
  - Made internal changes to `ParquetFile`, `ParquetRowReader` and `ParquetRowWriter` to allow setting the factories
  - Added a unit test for new feature

**Note:** Currently, argument names must be explicitly mentioned when using the logical factories for writing:
`using var writer = ParquetFile.CreateRowWriter<TTupleWrite>(outputStream, columnNames: columnNames, logicalTypeFactory: logicalWriteTypeFactory, logicalWriteConverterFactory: logicalWriteConverterFactory);`

I wasn't sure how to implement the overloads in such a way that specifying the argument names isn't necessary.

I'm happy to implement any changes or suggestions you see fit to improve the readability and maintainability of the feature!

### DIFF
--- a/csharp.test/TestRowOrientedParquetFile.cs
+++ b/csharp.test/TestRowOrientedParquetFile.cs
@@ -67,6 +67,17 @@ namespace ParquetSharp.Test
         }
 
         [Test]
+        public static void TestCustomTypeRoundtrip()
+        {
+            TestRoundtrip(new[]
+            {
+                new Row3 {A = 123, B = new VolumeInDollars(3.14f)},
+                new Row3 {A = 456, B = new VolumeInDollars(1.27f)},
+                new Row3 {A = 789, B = new VolumeInDollars(6.66f)}
+            });
+        }
+
+        [Test]
         public static void TestMappedToColumnAttributeOnRead()
         {
             TestRoundtripMapped<Row1, MappedRow1>(new[]
@@ -403,6 +414,11 @@ namespace ParquetSharp.Test
             RoundTripAndCompare(rows, rows, columnNames);
         }
 
+        private static void TestCustomTypeRoundtrip<TTuple>(TTuple[] rows)
+        {
+            CustomTypeRoundTripAndCompare(rows, rows, columnNames: null, new LogicalReadConverterFactory(), new LogicalWriteConverterFactory());
+        }
+
         private static void TestRoundtripMapped<TTupleWrite, TTupleRead>(TTupleWrite[] rows)
         {
             var expectedRows = rows.Select(
@@ -425,6 +441,24 @@ namespace ParquetSharp.Test
 
             using var inputStream = new BufferReader(buffer);
             using var reader = ParquetFile.CreateRowReader<TTupleRead>(inputStream);
+
+            var values = reader.ReadRows(rowGroup: 0);
+            Assert.AreEqual(expectedRows, values);
+        }
+
+        private static void CustomTypeRoundTripAndCompare<TTupleWrite, TTupleRead>(TTupleWrite[] rows, IEnumerable<TTupleRead> expectedRows, string[]? columnNames, LogicalReadConverterFactory logicalReadConverterFactory, LogicalWriteConverterFactory logicalWriteConverterFactory) {
+            using var buffer = new ResizableBuffer();
+
+            using (var outputStream = new BufferOutputStream(buffer))
+            {
+                using var writer = ParquetFile.CreateRowWriter<TTupleWrite>(outputStream, columnNames, logicalWriteConverterFactory);
+
+                writer.WriteRows(rows);
+                writer.Close();
+            }
+
+            using var inputStream = new BufferReader(buffer);
+            using var reader = ParquetFile.CreateRowReader<TTupleRead>(inputStream, logicalReadConverterFactory);
 
             var values = reader.ReadRows(rowGroup: 0);
             Assert.AreEqual(expectedRows, values);
@@ -470,10 +504,11 @@ namespace ParquetSharp.Test
         }
 
         [StructLayout(LayoutKind.Sequential)]
-        private readonly struct VolumeInDollars
+        private readonly struct VolumeInDollars : IEquatable<VolumeInDollars>
         {
             public VolumeInDollars(float value) { Value = value; }
             public readonly float Value;
+            public bool Equals(VolumeInDollars other) => Value.Equals(other.Value);
         }
 
         private struct MappedRow1
--- a/csharp.test/TestRowOrientedParquetFile.cs
+++ b/csharp.test/TestRowOrientedParquetFile.cs
@@ -451,7 +451,7 @@ namespace ParquetSharp.Test
             var logicalReadConverterFactory = new ReadConverterFactory();
             var logicalWriteConverterFactory = new WriteConverterFactory();
             var logicalWriteTypeFactory = new WriteTypeFactory();
-            var logicalReadTypeFactory = new ReadTypeFactoryNoOverride();
+            var logicalReadTypeFactory = new ReadTypeFactory();
 
             using (var outputStream = new BufferOutputStream(buffer))
             {
@@ -518,14 +518,14 @@ namespace ParquetSharp.Test
         /// <summary>
         /// A logical type factory that supports our user custom type (for the read tests only). Ignore overrides (used by unit tests that cannot provide a columnLogicalTypeOverride).
         /// </summary>
-        private sealed class ReadTypeFactoryNoOverride : LogicalTypeFactory
+        private sealed class ReadTypeFactory : LogicalTypeFactory
         {
             public override (Type physicalType, Type logicalType) GetSystemTypes(ColumnDescriptor descriptor, Type? columnLogicalTypeOverride)
             {
                 // We have to use the column name to know what type to expose.
                 Assert.IsNull(columnLogicalTypeOverride);
                 using var descriptorPath = descriptor.Path;
-                return base.GetSystemTypes(descriptor, descriptorPath.ToDotVector().First() == "values" ? typeof(VolumeInDollars) : null);
+                return base.GetSystemTypes(descriptor, descriptorPath.ToDotVector().First() == "B" ? typeof(VolumeInDollars) : null);
             }
         }
 
--- a/csharp.test/TestRowOrientedParquetFile.cs
+++ b/csharp.test/TestRowOrientedParquetFile.cs
@@ -461,7 +461,7 @@ namespace ParquetSharp.Test
             }
 
             using var inputStream = new BufferReader(buffer);
-            using var reader = ParquetFile.CreateRowReader<TTupleRead>(inputStream, logicalReadConverterFactory);
+            using var reader = ParquetFile.CreateRowReader<TTupleRead>(inputStream, logicalTypeFactory: logicalTypeFactory, logicalReadConverterFactory: logicalReadConverterFactory);
 
             var values = reader.ReadRows(rowGroup: 0);
             Assert.AreEqual(expectedRows, values);
--- a/csharp.test/TestRowOrientedParquetFile.cs
+++ b/csharp.test/TestRowOrientedParquetFile.cs
@@ -6,6 +6,7 @@ using ParquetSharp.IO;
 using ParquetSharp.RowOriented;
 using NUnit.Framework;
 using System.Linq;
+using System.Runtime.InteropServices;
 
 #if DUMP_EXPRESSION_TREES
 using System.Linq.Expressions;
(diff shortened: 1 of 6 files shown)


csharp/RowOriented/ParquetRowReader.cs
```diff
@@ -13,29 +13,30 @@ public sealed class ParquetRowReader<TTuple> : IDisposable
     {
         internal delegate void ReadAction(ParquetRowReader<TTuple> parquetRowReader, TTuple[] rows, int length);
 
-        internal ParquetRowReader(string path, ReadAction readAction, MappedField[] fields)
-            : this(new ParquetFileReader(path), readAction, fields)
+        internal ParquetRowReader(string path, ReadAction readAction, MappedField[] fields, LogicalTypeFactory? logicalTypeFactory = null, LogicalReadConverterFactory? logicalReadConverterFactory = null)
```
adamreeve, line 16: I think the issue might be that the `logicalTypeFactory` isn't being passed through to the underlying `ParquetFileReader`. The constructors here all ignore that parameter.

There's a comment in the docs that matches with the behaviour you describe: https://github.com/G-Research/ParquetSharp/blob/7d8b7bc995f072ab11e829819368fa5806289080/docs/TypeFactories.md?plain=1#L15

csharp/RowOriented/ParquetFile.cs
```diff
@@ -20,35 +20,49 @@ public static class ParquetFile
         /// <summary>
         /// Create a row-oriented reader from a file.
         /// </summary>
-        public static ParquetRowReader<TTuple> CreateRowReader<TTuple>(string path)
+        public static ParquetRowReader<TTuple> CreateRowReader<TTuple>(
+            string path,
+            LogicalTypeFactory? logicalTypeFactory = null,
+            LogicalReadConverterFactory? logicalReadConverterFactory = null)
```
adamreeve, line 26: I haven't taken a good look at all of your code changes but have one comment here. We try to avoid breaking binary compatibility so that users can use a new version of ParquetSharp in an application built against an older version. This means that rather than adding new optional parameters to existing public methods, you'll need to keep the existing methods as they are and add new overloads that take additional parameters.  There's a good article describing breaking changes in .NET libraries at https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes

csharp.test/TestRowOrientedParquetFile.cs
```diff
@@ -456,6 +494,71 @@ private struct Row2
...
+                // VolumeInDollars is bitwise identical to float, so we can reuse the native converter.
+                if (typeof(TLogical) == typeof(VolumeInDollars)) return LogicalRead.GetNativeConverter<VolumeInDollars, float>();
+                return base.GetConverter<TLogical, TPhysical>(columnDescriptor, columnChunkMetaData);
+            }
+        }
+
+        /// <summary>
+        /// A logical type factory that supports our user custom type (for the write tests only). Rely on overrides (used by unit tests that can provide a columnLogicalTypeOverride).
+        /// </summary>
+        private sealed class WriteTypeFactory : LogicalTypeFactory
```
adamreeve, line 541: Another issue is you use this type factory for both the read and write operations, but currently it only overrides `TryGetParquetTypes` which is used for writing. You'll need to implement `GetSystemTypes` for this to work for reading too.

csharp/RowOriented/ParquetFile.cs
jescalada: Makes sense, thank you!

I'll work on rewriting the code as new methods instead. Would you say it's necessary to add an extra overload for each existing method `ParquetFile.CreateRowReader()` and `ParquetFile.CreateRowWriter()`, along with the `ParquetRowReader/Writer` constructors?

I am currently setting the relevant factories in the common constructors for `ParquetRowReader` and `ParquetRowWriter`. Would you say this is the right way to go?

csharp.test/TestRowOrientedParquetFile.cs
jescalada: Oh, I didn't notice that, thank you!

csharp/RowOriented/ParquetRowReader.cs
jescalada: I'm a bit confused about this. I noticed that for the writing scenario, the `ParquetRowWriter` constructors all admit the `LogicalTypeFactory` as a parameter to the `ParquetFileWriter`, therefore the common constructor only requires a `LogicalWriteConverterFactory` to work. On the other hand, the `ParquetFileReader` doesn't admit a `LogicalTypeFactory`. Does this mean I need to set both in the common constructor for reading?

### ParquetRowWriter.cs
image

### ParquetRowReader.cs
image

jescalada:
It seems I'm still getting the same exception as before, despite implementing the fixes for the Read side of things...

I wonder if it's because I didn't write the Parquet file properly to begin with? Is there an easy way to debug the contents of the Parquet file as the test is executing?

adamreeve:
> It seems I'm still getting the same exception as before, despite implementing the fixes for the Read side of things...
> 
> I wonder if it's because I didn't write the Parquet file properly to begin with? Is there an easy way to debug the contents of the Parquet file as the test is executing?

I haven't looked closely at your latest changes, but my changes here on top of your original changes were enough to get the test to pass: https://github.com/adamreeve/ParquetSharp/commit/7767e7b3a890e45a6338a30411c24b4cf8ff07d9

jescalada:
Oh, I noticed that I forgot to change the column name for the ReadTypeFactory!
`return base.GetSystemTypes(descriptor, descriptorPath.ToDotVector().First() == "B" ? typeof(VolumeInDollars) : null);`

I'll add the new overloaded factory methods and get the PR out ASAP.

Thanks for the help! :)

csharp/RowOriented/ParquetRowReader.cs
adamreeve: I think this is because the writer requires the logical type factory to be present at construction time to convert from the `Column` types to the schema to write, but the reader only needs it when you start reading data, so doesn't take it in the constructor. But you will still need to set both for the mapping to work to correctly.

csharp/RowOriented/ParquetFile.cs
adamreeve: I don't think you need to add extra overloads corresponding to each existing method, it might be sufficient to just add one new overload that takes all of the other possible parameters too. Although there might be some cases where you need multiple overloads to accept different input types, eg. reading a stream vs a string for a file path. And yes, setting these in the `ParquetRowReader` and `ParquetRowWriter` looks like a good approach.

adamreeve:
> Oh, I noticed that I forgot to change the column name for the ReadTypeFactory! `return base.GetSystemTypes(descriptor, descriptorPath.ToDotVector().First() == "B" ? typeof(VolumeInDollars) : null);`

Ah right, good catch. It is a little awkward that the column name needs to be used there. Possibly we can improve this in the future as when we use the row oriented API we know the types we expect for each column so should be able to be a bit smarter, eg. maybe by overriding the type factory automatically.

But I think it's fine for now to need to use the column names in the type factory. The original feature request was only about writing, and we can add that as a follow up feature later if we want. It would be good to document how to use this feature though, and the requirement to match columns on their names for mapping to types for reading.

adamreeve:
By the way, to get the format check to pass in CI you can run the formatter tool with:
```
dotnet tool restore
```

csharp/RowOriented/ParquetFile.cs
jescalada: I'm currently trying to do this with a single overload, however I can't find a way other than checking if one of the parameters is null (say, the `ReaderProperties`) in order to pick a constructor.

I thought perhaps it might be alright to create the additional overloads, because the underlying constructors for `ParquetRowReader`/`ParquetRowWriter` cannot be removed due to the API compatibility/breaking changes issue (suppose if I added a general overload that makes the previous constructors obsolete).

csharp/RowOriented/ParquetRowWriter.cs
```diff
@@ -78,6 +135,8 @@ public void Close()
         public ColumnDescriptor ColumnDescriptor(int i) => _parquetFileWriter.ColumnDescriptor(i);
         public FileMetaData? FileMetaData => _parquetFileWriter.FileMetaData;
         public IReadOnlyDictionary<string, string> KeyValueMetadata => _parquetFileWriter.KeyValueMetadata;
+        public LogicalTypeFactory LogicalTypeFactory { get; set; } = LogicalTypeFactory.Default;
+        public LogicalWriteConverterFactory LogicalWriteConverterFactory { get; set; } = LogicalWriteConverterFactory.Default;
```
adamreeve, line 139: These should be removed too

csharp/RowOriented/ParquetRowWriter.cs
```diff
@@ -61,6 +109,15 @@ private ParquetRowWriter(ParquetFileWriter parquetFileWriter, WriteAction writeA
             _rows = new TTuple[1024];
         }
 
+        private ParquetRowWriter(ParquetFileWriter parquetFileWriter, WriteAction writeAction, LogicalWriteConverterFactory? logicalWriteConverterFactory = null)
```
adamreeve, line 112: I think you can remove the original private constructor above and just keep this one. Because it's private there's no concern about binary compatibility here

csharp/RowOriented/ParquetRowWriter.cs
```diff
@@ -61,6 +109,15 @@ private ParquetRowWriter(ParquetFileWriter parquetFileWriter, WriteAction writeA
...
+        {
+            _parquetFileWriter = parquetFileWriter;
+            _parquetFileWriter.LogicalWriteConverterFactory = logicalWriteConverterFactory ?? LogicalWriteConverterFactory.Default;
```
adamreeve, line 115: Nitpick but I think it's probably clearer to only assign to the converter factory if it's non-null, as it should already be set to the default factory:
```suggestion
            if (logicalWriteConverterFactory != null)
            {
                _parquetFileWriter.LogicalWriteConverterFactory = logicalWriteConverterFactory;
            }
```

csharp/RowOriented/ParquetFile.cs
adamreeve: Okay yes it's fine to add the extra overloads then if that's the simplest approach. Although the constructors for `ParquetRowReader` and `ParquetRowWriter` are internal only, so you can change their signature if needed. It's only the public factory methods (`CreateRowReader`/`CreateRowWriter`) on `ParquetFile` that need to be kept compatible.

adamreeve (commented):
This is looking pretty good thanks @jescalada! I've left a few minor comments. It would also be great if you could add a small section to the row-oriented API docs to describe how to use this feature.

csharp/RowOriented/ParquetRowWriter.cs
jescalada: I think I understand how binary compatibility works now, thanks!

csharp/RowOriented/ParquetRowWriter.cs
adamreeve: :+1: yep having separate checks makes sense there

docs/RowOriented.md
```diff
@@ -68,6 +68,120 @@ using (var rowReader = ParquetFile.CreateRowReader<MyRow>("example.parquet"))
 }
 ```
 
+## Reading and writing custom types
+
+The row-oriented API supports reading and writing custom types by providing a `LogicalReadConverterFactory` or `LogicalWriteConverterFactory`.
+The `LogicalTypeFactory` is required if the custom type is used for creating the schema (when writing), or if accessing a `LogicalColumnReader`
```
adamreeve, line 74: When using the row-oriented API, I believe the `LogicalTypeFactory` will always be required as you can't access the lower level column based API, so it might be simplest to just say both the converter factory and type factory are required without going into further detail here.

adamreeve (commented):
This looks good thanks @jescalada. Just one minor comment on the docs and then I think this can be merged after we've finalised the 13.0.0 release.

docs/RowOriented.md
jescalada: Thank you @adamreeve! 😄

docs/RowOriented.md
```diff
@@ -68,6 +68,117 @@ using (var rowReader = ParquetFile.CreateRowReader<MyRow>("example.parquet"))
 }
 ```
 
+## Reading and writing custom types
+
+The row-oriented API supports reading and writing custom types by providing a `LogicalReadConverterFactory` or `LogicalWriteConverterFactory`.
```
adamreeve, line 73: We should still mention the `LogicalTypeFactory` here:
```suggestion
The row-oriented API supports reading and writing custom types by providing a `LogicalTypeFactory` and a `LogicalReadConverterFactory` or `LogicalWriteConverterFactory`.
```

jescalada: That makes more sense :)

adamreeve (approved):

marcin-krystianc (approved):

PR #397, opened 2023-11-23
Small bit of simplification

philjdf:
1. Noticed an unused parameter `elementType` in the constructor of `LogicalColumnStream`, and also that the `Buffer` etc members of that class are unused except at the leaf subclass levels, so this looks like it should be composition rather than inheritance. 
2. Noticed `LogicalColumnReader<TPhysical, TLogical, TElement>` has `TPhysical` and `TLogical` generic type args only being used at construction time, so move those types to a static factory method. Also rename it to `LogicalColumnReaderBatch` to avoid the clash with the base class.
3. Now there doesn't seem much point in having `LogicalColumnReaderBatch`, merge it with its parent class `LogicalColumnReader`. So instead of having `sealed LogicalColumnReader<TPhysical, TLogical, TElement> : abstract LogicalColumnReader<TElement> : abstract LogicalColumnReader` we've got `sealed LogicalColumnReader<TElement> : abstract LogicalColumnReader`.
4. Same on the `LogicalColumnWriter` side.
5. Move the `LogicalColumnStreamBuffer` creation inside the `LogicalBatchReader/WriterFactory` classes, and it turns out there's already a `LogicalStreamBuffers<TPhysical>` class there which we can use instead.


adamreeve (approved):
This looks great thanks Phil!

PR #398, opened 2023-11-25
Document potential build issues in README

jescalada:
Hi, this is a PR for some more documentation enhancements related to potential issues users may encounter when building locally.

I went through the manual build process myself and I hit quite a few roadblocks which might cause friction for new users, especially relating to having all the available Visual Studio modules installed in order to run the build script.

This is not a comprehensive troubleshooting guide, but it aims to solve some of the errors that may come up after running the script (which may not entirely obvious, as they do occur even after installing the required C++ tools).

Please let me know if there's any other changes you would like to see, or if you would prefer if this troubleshooting section were in another file in the `docs`, rather than to expand the README.

## Changelog
- Added troubleshooting guides for various build issues
  - `CMake Error: Visual Studio not found`
    - Apparently a bug
    - Solution is to reinstall Build Tools and reboot or reinstall Visual Studio
  - `Microsoft.Cpp.Default.props not found`
    - Due to missing Build Tools module (not installed by default)
  - `CMake Error: Compiler not found`
    - Solution is to install missing module (which is not installed by default in "Desktop Development with C++" module)

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,8 @@ build_windows.ps1
 dotnet test csharp.test
 ```
 
+`cmake` must be available in the PATH for the build script to succeed.
+
 #### Unix
 
 Build the C++ code and run the C# tests with:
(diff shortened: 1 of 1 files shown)


adamreeve (approved):
:+1: thanks @jescalada

PR #401, opened 2023-11-26
Fix segfault writing to no longer valid column writer

### LINKED ISSUE #400: [BUG]: Segfault when writing to a column writer after getting the next column writer
### Issue Description

Writing to a column writer after calling `NextColumn` on the `RowGroupWriter` again can cause a crash and exit with code 139.

### Environment Information

- ParquetSharp Version:  12.0.1
- .NET Framework/SDK Version: .NET 7
- Operating System: Fedora 38

### Steps To Reproduce

```c#
var schemaColumns = new Column[]
{
    new Column<int>("A"),
    new Column<float>("B"),
};

using var outStream = new BufferOutputStream(buffer);
using var writer = new ParquetFileWriter(outStream, schemaColumns);
using var rowGroupWriter = writer.AppendRowGroup();

using var colWriterA = rowGroupWriter.NextColumn().LogicalWriter<int>();
using var colWriterB = rowGroupWriter.NextColumn().LogicalWriter<float>();

colWriterA.WriteBatch(new[] {0, 1, 2, 3, 4});

writer.Close();
```

### Expected Behavior

This should throw a .NET exception rather than a segfault that can't be handled.

### Additional Context (Optional)

Relevant documentation on the C++ `NextColumn` method is here: https://github.com/apache/arrow/blob/6af660f48472b8b45a5e01b7136b9b040b185eb1/cpp/src/parquet/file_writer.h#L67-L74

When fixing this, we'll need to account for buffered row groups where we can write to columns in any order.

adamreeve:
Fixes #400

### DIFF
--- a/csharp/ColumnWriter.cs
+++ b/csharp/ColumnWriter.cs
@@ -163,11 +163,14 @@ namespace ParquetSharp
         {
             get
             {
-                var currentColumn = RowGroupWriter.CurrentColumn;
-                if (ColumnIndex != currentColumn)
+                if (!RowGroupWriter.Buffered)
                 {
-                    throw new Exception($"Writer for column {ColumnIndex} is no longer valid, " +
-                                        $"the current column for the row group writer is {currentColumn}");
+                    var currentColumn = RowGroupWriter.CurrentColumn;
+                    if (ColumnIndex != currentColumn)
+                    {
+                        throw new Exception($"Writer for column {ColumnIndex} is no longer valid, " +
+                                            $"the current column for the row group writer is {currentColumn}");
+                    }
                 }
 
                 return _handle;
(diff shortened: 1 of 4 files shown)


csharp/ColumnWriter.cs
```diff
@@ -159,7 +159,25 @@ public LogicalColumnWriter<TElement> LogicalWriterOverride<TElement>(int bufferL
         protected static extern unsafe IntPtr TypedColumnWriter_WriteBatchSpaced_FixedLenByteArray(
             IntPtr columnWriter, long numValues, short* defLevels, short* repLevels, byte* validBits, long validBitsOffset, FixedLenByteArray* values);
 
-        protected readonly IntPtr Handle;
+        protected IntPtr Handle
```
adamreeve, line 162: I don't believe we need to worry about binary compatibility here, because `Handle` is a protected member and the `ColumnWriter` and `ColumnWriter<TValue>` constructors are both internal, so it shouldn't be possible for code outside of ParquetSharp to be accessing this member.

PR #408, opened 2024-01-09
Copy vcpkg.json before cloning vcpkg in devcontainer Dockerfile

adamreeve:
After upgrading to Arrow 14.0.2 the devcontainer build failed due to not finding that Arrow version in vcpkg (https://github.com/G-Research/ParquetSharp/actions/runs/7458957587/job/20293957096), although the devcontainer build had passed in the PR.

It looks like this is because the docker build used a cached result of checking out the vcpkg repository from before Arrow 14.0.2 was added.

I've moved the copy of vcpkg.json to before the vcpkg repo checkout, so any changes to vcpkg.json should force a new checkout of vcpkg rather than use the previously cached result.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -46,15 +46,15 @@ FROM cpp AS vcpkg
 
 USER vscode
 
+# Copy our vcpkg manifest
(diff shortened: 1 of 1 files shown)


marcin-krystianc (approved):
Good catch 👍

PR #414, opened 2024-01-31
Lint: Update rules for formating code

ljubon:
Based on existing rule sets in [ParquetSharp.DotSettings](https://github.com/ljubon/ParquetSharp/blob/oss-414-linting-lvl-4/ParquetSharp.DotSettings) and as part of [OSS-414](https://github.com/G-Research/oss-portfolio-maturity/issues/414) goal of this PR is to add new rules which are safe and not breaking current process, just enforcing existing syntax and style

_Optional_: I tried one more rule to add, but it's not part of this PR
Here are [changes ](https://github.com/ljubon/ParquetSharp/actions/runs/7506069064/job/20436665643#step:4:3173), total `67` file
Rule line is:
```xml
<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/SPACE_AFTER_TYPECAST_PARENTHESES/@EntryValue">False</s:Boolean>
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Setup .NET SDK v7.0.x
         uses: actions/setup-dotnet@v3
         with:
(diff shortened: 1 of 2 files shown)


marcin-krystianc (dismissed):

adamreeve:
I'm not sure about the `SPACE_AFTER_TYPECAST_PARENTHESES` change. I prefer the space there, but it looks like no space is the default and the standard that a lot of C# projects follow.

I don't feel too strongly either way, but if we want to keep the spaces, then we should add that check but set it to true rather than false.

adamreeve:
Yes that would be helpful thanks, although I guess you mean set it to true?

adamreeve:
Cool, thanks @ljubon, I'm happy with that

adamreeve (approved):

PR #418, opened 2024-02-04
Add support for reading and writing the .NET Half type

### LINKED ISSUE #413: Support round-tripping Half values using the new Float16 logical type
Arrow 15 adds support for a new [Float16 logical type](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#float16) in Parquet

We should investigate supporting round-tripping .NET `Half` values with this type, although this is only available since .NET 5 so may require adding a new target.

adamreeve:
Fixes #413 

This adds support for the new `Float16` logical type added in Arrow 15. I've added a new .NET 6 target to the ParquetSharp project to allow using the new `Half` type, which required fixing a few errors related to nullable reference type checking when building with the newer target.

One thing to be aware of is that writing `Half` values can be a bit slower than floats because of the extra overhead of writing these as fixed-length byte arrays rather than having a dedicated physical type. We could possibly improve this in future if it turns out to be a problem. I did some quick benchmarking of reading and writing 1 million random floats, doubles and half values, with dictionary encoding disabled:

|     Method |     Mean |     Error |    StdDev | Ratio | RatioSD |
|----------- |---------:|----------:|----------:|------:|--------:|
|   ReadHalf | 1.267 ms | 0.0253 ms | 0.0329 ms |  1.02 |    0.02 |
|  ReadFloat | 1.240 ms | 0.0236 ms | 0.0299 ms |  1.00 |    0.00 |
| ReadDouble | 2.292 ms | 0.0411 ms | 0.0563 ms |  1.85 |    0.07 | 

|      Method |      Mean |     Error |    StdDev |    Median | Ratio | RatioSD |
|------------ |----------:|----------:|----------:|----------:|------:|--------:|
|   WriteHalf | 16.771 ms | 0.3318 ms | 0.5452 ms | 16.466 ms |  3.13 |    0.25 |
|  WriteFloat |  5.336 ms | 0.1280 ms | 0.3775 ms |  5.324 ms |  1.00 |    0.00 |
| WriteDouble |  8.845 ms | 0.1759 ms | 0.2408 ms |  8.849 ms |  1.64 |    0.11 |

### DIFF
--- a/csharp/LogicalRead.cs
+++ b/csharp/LogicalRead.cs
@@ -664,7 +664,20 @@ namespace ParquetSharp
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static unsafe Half ToHalf(FixedLenByteArray source)
         {
-            return *(Half*) source.Pointer;
+            if (BitConverter.IsLittleEndian)
+            {
+                return *(Half*) source.Pointer;
+            }
+            else
+            {
+                // Float-16 values are always stored in little-endian order
+                var value = (Half) 0.0f;
+                var dest = (byte*) &value;
+                var src = (byte*) source.Pointer;
+                dest[1] = src[0];
+                dest[0] = src[1];
+                return value;
+            }
         }
 #endif
 
--- a/csharp/LogicalRead.cs
+++ b/csharp/LogicalRead.cs
@@ -167,6 +167,18 @@ namespace ParquetSharp
                 return (LogicalRead<Guid?, FixedLenByteArray>.Converter) LogicalRead.ConvertUuid;
             }
 
+#if NET5_0_OR_GREATER
+            if (typeof(TLogical) == typeof(Half))
+            {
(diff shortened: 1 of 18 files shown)


csharp/LogicalRead.cs
```diff
@@ -630,6 +660,27 @@ public static unsafe Guid ToUuid(FixedLenByteArray source)
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe Half ToHalf(FixedLenByteArray source)
+        {
+            if (BitConverter.IsLittleEndian)
+            {
+                return *(Half*) source.Pointer;
+            }
+            else
+            {
+                // Float-16 values are always stored in little-endian order
```
adamreeve, line 673: For completeness I thought we should handle big-endian machines, as we have a similar check for the guid type, but we currently only build x64 and arm64 native binaries for the nuget package, and there doesn't seem to be an easy way to test this. Maybe we should just throw a not implemented exception for big endian instead?

marcin-krystianc: I think it is fine as it is now.

PR #426, opened 2024-02-28
Add the key management tools API for encryption

### LINKED ISSUE #200: Integrate high level encryption API
See https://github.com/apache/arrow/pull/9631 for discussions and details.

Probably we want to expose these as a new namespace to mirror the C++ API. Not clear how best to integrate this with ParquetSharp, since the new API is still ultimately converted to the low-level API before being used by a reader/writer.

adamreeve:
Fixes #200

This adds bindings to the Parquet key management tools API, also known as the high level encryption API. The existing low-level encryption API  supports setting encryption/decryption keys directly, whereas this API uses a Key Management System (KMS) to implement envelope encryption, where data encryption keys are randomly generated and then stored encrypted in the file metadata.


csharp/Encryption/CryptoFactory.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+
+        /// <summary>
+        /// Get decryption properties for a Parquet file.
+        /// If external key material is used then the path to the parquet file must be provided.
+        /// This CryptoFactory instance must remain alive and not disposed until after any files using these
+        /// decryption properties have been read, as internally the FileDecryptionProperties contains references to
+        /// data in the CryptoFactory that cannot be managed by ParquetSharp.
+        /// Failure to do so may result in native memory access violations and crashes that cannot be caught as exceptions.
```
adamreeve, line 57: This is a bit awkward. I couldn't find a way to handle this nicely without exposing users to possible segfaults, as the `FileDecryptionProperties` class in C++ has a private constructor and can't be subclassed, so we can't do anything like combine it with a reference to the CryptoFactory. I'm hoping this can be fixed in Arrow so we can fix this limitation for the next version though.

adamreeve:
I'm not sure what to do about the Windows build failure. It worked for me with MSVC 19.35.32216.1, and with 19.39.33521.0 after upgrading. The GitHub runner has version 19.38.33135.0 though.

It looks like this is triggered by my changes pulling in a dependency on the Arrow FileSystem API, but this seems to be a bug in MSVC. I found a similar issue reported at https://developercommunity.visualstudio.com/t/Latest-Visual-Studio-version-1791-ca/10598722

We might need to wait until the runners get a newer version of MSVC, or disable this feature on Windows for now.

adamreeve:
Similar build failure issue reported at https://github.com/actions/runner-images/issues/9398, but that was apparently caused by linking to a different STL version than what the compiler expected. Maybe there's something similar going on with our build.

adamreeve:
The build failure appears to be because Arrow is built with a different MSVC than ParquetSharp, which isn't good. I'm not sure why different versions are being selected:

From the Arrow build logs:
```
-- The CXX compiler identification is MSVC 19.39.33520.0
...
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.39.33519/bin/Hostx64/x64/cl.exe - skipped
```

And from the ParquetSharp build:
```
-- The CXX compiler identification is MSVC 19.38.33135.0
...
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.38.33130/bin/HostX64/x64/cl.exe - skipped
```

marcin-krystianc (dismissed):
Amazing work, well done!

csharp/Encryption/EncryptionConfiguration.cs
```diff
@@ -0,0 +1,227 @@
...
+        /// management tools. Can be 128, 192 or 256 bits.
+        /// The default is 128 bits.
+        /// </summary>
+        public int DataKeyLengthBits
+        {
+            get => ExceptionInfo.Return<int>(Handle, EncryptionConfiguration_GetDataKeyLengthBits);
+            set => ExceptionInfo.Check(EncryptionConfiguration_SetDataKeyLengthBits(Handle.IntPtr, value));
+        }
+
+        private static string EncodeColumnKeys(IReadOnlyDictionary<string, IReadOnlyList<string>> columnKeys)
```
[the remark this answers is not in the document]
adamreeve: Good point, that's much more readable, I've changed this now

csharp/Encryption/CryptoFactory.cs
adamreeve: I've made a PR in Arrow to address this at https://github.com/apache/arrow/pull/40329

marcin-krystianc (approved):

csharp/Encryption/CryptoFactory.cs
adamreeve: That's been merged, so we should be able to remove this warning in the 16.0.0 release

PR #430, opened 2024-03-19
Allow getting a ParquetFileReader and the SchemaManifest from an Arrow.FileReader

adamreeve:
* This allows getting a `ParquetFileReader` from an `Arrow.FileReader`, which might be required if you want to access things like the row group statistics
* In order to allow this in a safe way, I've added a new `ChildParquetHandle` class that holds a pointer to an object that is owned by some other disposable object, and added a new `INativeHandle` interface to support using either a `ParquetHandle` or `ChildParquetHandle`. There are other places in the code where we might want to use this, eg. the `RowGroupMetaData` returned from a `RowGroupReader` is one I noticed, but that should probably be a separate PR. See also #322
* Additionally, to be able to interpret the Parquet metadata when the file has a complex nested structure, expose the `SchemaManifest` associated with a `FileReader`

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -67,6 +67,8 @@ add_library(ParquetSharpNative SHARED
 	arrow/ArrowWriterPropertiesBuilder.cpp
 	arrow/FileReader.cpp
 	arrow/FileWriter.cpp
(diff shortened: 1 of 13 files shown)


marcin-krystianc (approved):
Nice 👍

PR #444, opened 2024-05-01
Fix breaking change to LogicalTypeEnum.None value

### LINKED ISSUE #443: [BUG]: 15.0.2 release breaks binary compatibility by changing the value of LogicalTypeEnum.None
### Issue Description

The 15.0.2 release changed the value of `LogicalTypeEnum.None` (#417), which means that any library or application compiled against an older version of ParquetSharp may behave incorrectly if it uses 15.0.2 at runtime.

### Environment Information

- ParquetSharp Version:  15.0.2
- .NET Framework/SDK Version: .NET 8.0.104
- Operating System: Fedora Linux 39

### Steps To Reproduce

Build the following code against ParquetSharp 14.0.2:
```c#
using var noneType = LogicalType.None();
Console.WriteLine(
    $"None logical type enum value = {noneType.Type} ({(int) noneType.Type}), " +
    $"expected {LogicalTypeEnum.None} ({(int) LogicalTypeEnum.None})");
```

Then run it using the ParquetSharp 15.0.2 binaries (`ParquetSharp.dll` and `ParquetSharpNative.so`). This outputs:

```
None logical type enum value = None (16), expected Float16 (15)
```

### Expected Behavior

The output should be the same as when using ParquetSharp 14.0.2 at runtime, which is:
```
None logical type enum value = None (15), expected None (15)
```

### Additional Context (Optional)

_No response_

adamreeve:
Fixes #443

### DIFF
--- a/csharp/LogicalType.cs
+++ b/csharp/LogicalType.cs
@@ -20,8 +20,62 @@ namespace ParquetSharp
         Json = 12,
         Bson = 13,
         Uuid = 14,
+        None = 15,
+        Float16 = 16,
+    }
+
+    /// <summary>
+    /// Type enum with values that match the C++ Parquet library.
+    /// These are not guaranteed to be stable between releases, but we need to keep
+    /// the LogicalTypeEnum values above stable to ensure ABI compatibility.
+    /// </summary>
+    internal enum CppLogicalTypeEnum
+    {
+        Undefined = 0,
+        String = 1,
+        Map = 2,
+        List = 3,
+        Enum = 4,
+        Decimal = 5,
+        Date = 6,
+        Time = 7,
+        Timestamp = 8,
+        Interval = 9,
+        Int = 10,
+        Nil = 11,
+        Json = 12,
+        Bson = 13,
+        Uuid = 14,
         Float16 = 15,
-        None = 16
+        None = 16,
+    }
+
+    internal static class CppLogicalTypeEnumExtensions
+    {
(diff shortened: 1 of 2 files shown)


marcin-krystianc (approved):
Writing this enum type converter seems to be an overkill, but on the other hand there is rather no other option. In large solutions, there is a high chance of having projects compiled against ParquetSharp pre-15.0.2 and 15.0.2, which would cause hard to debug errors at runtime.

PS:
What an odd decision to put the `None` as the lat value in the enum. It guarantees that each time a new value is added, the value of `None` needs to be changed as well.

marcin-krystianc (commented):
Should we implement additional measures to prevent from such mistakes in the future?

adamreeve:
> PS: What an odd decision to put the `None` as the lat value in the enum. It guarantees that each time a new value is added, the value of `None` needs to be changed as well.

Yeah, this is a little unfortunate and if we'd foreseen this problem we could maybe have mapped `None` to something different to what the C++ library uses that wasn't going to clash with a future value.

> Should we implement additional measures to prevent such mistakes in the future?

This is a good idea, we have some checks that detect if the C++ enum values change (https://github.com/G-Research/ParquetSharp/blob/9fba13153c272627db4438d064ce829198cda752/cpp/Enums.cpp#L40) but I only thought that meant the C# enums needed updating, I missed that this would break the ABI. Maybe we could have some automated tests that build tests against older versions of ParquetSharp and then run them with the current build, which should detect a range of different ABI breaking scenarios.

PR #446, opened 2024-05-03
Run tests built against the previous stable release in CI

adamreeve:
This adds a CI job to run the tests built against the previous stable release with the current dll and native library at runtime, to try to detect binary compatibility issues like #443

I've also updated the logical type roundtrip tests so that they would catch the logical type enum values changing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -297,6 +297,51 @@ jobs:
     - name: Build & Run F# unit tests
       run: dotnet test fsharp.test --configuration=Release --framework ${{ matrix.dotnet }}
 
+  test-binary-compatibility:
+    name: Test binary compatibility against the last stable release
+    runs-on: ubuntu-20.04
+    needs: build-nuget
+    env:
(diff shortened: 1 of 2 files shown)


marcin-krystianc (commented):
I think it is useful, but not ideal.
We are going to use the current test project to run tests against an old binary.
It is going to be a common situation when that is going to fail, because anytime a new feature with accompanying tests is implemented it is not going to work with an older version. 

How about we try to checkout previous version (previous commit) to run the tests against current binary (binary built from the current commit)?

marcin-krystianc:
> > How about we try to checkout previous version (previous commit) to run the tests against current binary (binary built from the current commit)?
> 
> This is the idea. `BASELINE_VERSION` is the old stable version to checkout and build the tests for, and we then run those against the new ParquetSharp.dll from the NuGet that has been built in previous steps.

Ah, I missed that!

adamreeve:
Closing as I don't believe this is necessary after #449.

PR #457, opened 2024-06-10
Use GitHub Actions for vcpkg binary caching

jgiannuzzi:
This PR switches our CI to use a more granular approach to vcpkg binary caching. It uses the [GitHub Actions Cache](https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-actions-cache) feature of vcpkg to cache individual packages instead of the all-or-nothing blob approach we were using. The huge advantage is that any package that gets properly built gets cached, even if the workflow does fail at some point in time. The other advantage is that we don't need to keep updating our cache keys to try to match the vcpkg ABI mechanism, but can instead let vcpkg do its thing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,38 +84,11 @@ jobs:
         echo "root=$VCPKG_INSTALLATION_ROOT" >> $GITHUB_OUTPUT
       shell: bash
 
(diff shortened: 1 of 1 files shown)


adamreeve (approved):
:+1: looks good thanks Jonathan

PR #466, opened 2024-06-24
Bump version of macos used to build the native library

adamreeve:
The macos-11 images are being deprecated, see failures due to the brownout in https://github.com/G-Research/ParquetSharp/actions/runs/9643175788.

I've just bumped to macos-12 rather than 13 or the ARM based macos-14 images as I'm not certain whether a native library built on macos would be compatible with older versions. It sounds like we might be able to work around that by setting the `MACOSX_DEPLOYMENT_TARGET` environment variable but I'm not familiar with building on mac.


adamreeve:
Closing, this is a dup of #465

PR #470, opened 2024-07-02
ci: use pypa/manylinux image instead of vanilla CentOS image

jgiannuzzi:
Following the EOL of CentOS 7, `mirrorlist.centos.org` no longer resolves. The [pypa/manylinux](https://github.com/pypa/manylinux) 2014 image has been updated to use `vault.centos.org` instead and will likely keep up with any other changes so I suggest we switch to it for now.


adamreeve (approved):
:+1: thanks Jonathan

PR #476, opened 2024-08-04
Add operators etc to DateTimeNanos

Brandon4BZ:


csharp/DateTimeNanos.cs
```diff
@@ -10,14 +10,37 @@ namespace ParquetSharp
     [StructLayout(LayoutKind.Sequential)]
     public readonly struct DateTimeNanos : IEquatable<DateTimeNanos>, IComparable, IComparable<DateTimeNanos>
     {
+        private const long DateTimeOffsetTicks = 621355968000000000; // new DateTime(1970, 01, 01).Ticks
+        private const string DefaultFormat = "yyyy-MM-dd HH:mm:ss.fffffffff";
```
adamreeve, line 14: Can you please leave private members at the bottom of the class definition for consistency with the ParquetSharp codebase.

csharp/PublicAPI/net471/PublicAPI.Shipped.txt
```diff
@@ -831,6 +831,13 @@ static ParquetSharp.Arrow.ArrowReaderProperties.GetDefault() -> ParquetSharp.Arr
 static ParquetSharp.Arrow.ArrowWriterProperties.GetDefault() -> ParquetSharp.Arrow.ArrowWriterProperties!
 static ParquetSharp.Column.CreateSchemaNode(ParquetSharp.Column![]! columns, ParquetSharp.LogicalTypeFactory! logicalTypeFactory, string! nodeName = "schema") -> ParquetSharp.Schema.GroupNode!
 static ParquetSharp.Column.CreateSchemaNode(ParquetSharp.Column![]! columns, string! nodeName = "schema") -> ParquetSharp.Schema.GroupNode!
+static ParquetSharp.DateTimeNanos.FromDotnetTicks(long dotnetTicks) -> ParquetSharp.DateTimeNanos
```
adamreeve, line 834: I think it's fine to leave these in `*.Shipped.txt` now, but the idea is they should go into the `*.Unshipped.txt` files and only be moved to `Shipped.txt` when a stable release is made.

adamreeve (changes requested):
Looks good thanks! Just a minor comment. And you'll need to run the formatter for ci to pass:
```
dotnet tool restore
```

csharp/PublicAPI/net471/PublicAPI.Shipped.txt
Brandon4BZ: Ah right - sorry! As you can see I haven't used this before ... will remember for next time!

adamreeve (approved):

csharp/DateTimeNanos.cs
```diff
@@ -89,18 +137,17 @@ public string ToString(string? format, IFormatProvider? formatProvider = null)
...
-        public static readonly DateTime MaxDateTimeValue = new DateTimeNanos(long.MaxValue).DateTime;
-
-        private const long DateTimeOffset = 621355968000000000; // new DateTime(1970, 01, 01).Ticks
-
+        private const long DateTimeOffsetTicks = 621355968000000000; // new DateTime(1970, 01, 01).Ticks
         private const string DefaultFormat = "yyyy-MM-dd HH:mm:ss.fffffffff";
+ 
+        private static long DotnetTicksToNanosSinceEpoch(long dotnetTicks)
+        {
+            return (dotnetTicks - DateTimeOffsetTicks) * 1_000_000L / TimeSpan.TicksPerMillisecond;
```
adamreeve, line 145: I was about to say you could change `private static readonly long NanosPerTick` to `private const long NanosPerTick` to fix the issue with it being zero when evaluating `MinDateTimeValue` and `MaxDateTimeValue`, which might be cleaner? If you want to expand it out here and below it would be good to keep the operation order the same as it was before to avoid potential overflow:
```suggestion
            return (dotnetTicks - DateTimeOffsetTicks) * (1_000_000L / TimeSpan.TicksPerMillisecond);
```

adamreeve (approved):

csharp/DateTimeNanos.cs
Brandon4BZ: Interesting - didn't know the compiler distinguished between constant and static readonly like that, thought I'd just made a silly mistake.

adamreeve:
Thanks for the PR @Brandon4BZ!

csharp/DateTimeNanos.cs
adamreeve: Yeah I didn't realise what had caused the test failures initially either, it makes sense now but didn't occur to me that the static member order would be important.

Brandon4BZ:
Thank you for the review! Sorry it was more of a faff than I had expected 😆

PR #482, opened 2024-08-27
Support reading and writing decimals with arbitrary length fixed-length byte array columns and writing with int32 or int64

### LINKED ISSUE #252: Support writing arbitrary precision decimal columns
Hello
I am trying to copy data from MS SQL table with NUMERIC(18,9) column "num" to Redshift with NUMERIC(18,9) column.
When I try to load data data into Redshift I get an error:
ERROR [XX000] [Amazon][Amazon Redshift] (30) Error occurred while trying to execute a query: [SQLState XX000] ERROR:  Spectrum Scan Error
DETAIL:
```
  error:  Spectrum Scan Error
  code:      15001
  context:   File 'https://s3.us-east-1.amazonaws.com/xyz/new_line.parquet' column 'num' has an invalid type length. Expecting: 8 len in file: 13
  query:     50511242
  location:  dory_util.cpp:1141
  process:   worker_thread [pid=12646]
```

When I change the destination column precision to 29, I was able to load data.
I would like to have destination table precision the same as source table.
Thanks,
Dmitriy

adamreeve:
Fixes #252

This adds a new `DecimalConverter` class that works with arbitrary length fixed-length byte array columns (as long as values are still representable by a .NET decimal), while keeping the existing `Decimal128` class for backwards compatibility and improved performance. I've also added support for writing decimal columns backed by the int32 and int64 physical types.

### DIFF
--- a/csharp/LogicalWrite.cs
+++ b/csharp/LogicalWrite.cs
@@ -101,12 +101,13 @@ namespace ParquetSharp
                 var multiplier = DecimalConverter.GetScaleMultiplier(columnDescriptor.TypeScale, columnDescriptor.TypePrecision);
                 if (typeof(TPhysical) == typeof(FixedLenByteArray))
                 {
-                    if (byteBuffer == null) throw new ArgumentNullException(nameof(byteBuffer));
-                    if (TypeUtils.UseDecimal128(columnDescriptor))
+                    if (byteBuffer == null)
                     {
-                        return (LogicalWrite<decimal, FixedLenByteArray>.Converter) ((s, _, d, _) => LogicalWrite.ConvertDecimal128(s, d, multiplier, byteBuffer));
+                        throw new ArgumentNullException(nameof(byteBuffer));
                     }
-                    return (LogicalWrite<decimal, FixedLenByteArray>.Converter) ((s, _, d, _) => LogicalWrite.ConvertDecimal(s, d, multiplier, byteBuffer, columnDescriptor.TypeLength));
+                    return TypeUtils.UseDecimal128(columnDescriptor)
+                        ? (LogicalWrite<decimal, FixedLenByteArray>.Converter) ((s, _, d, _) => LogicalWrite.ConvertDecimal128(s, d, multiplier, byteBuffer))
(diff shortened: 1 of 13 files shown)


csharp/LogicalWrite.cs
```diff
@@ -449,6 +479,85 @@ public static void ConvertDecimal128(ReadOnlySpan<decimal?> source, Span<short>
             }
         }
 
+        public static void ConvertDecimal(ReadOnlySpan<decimal> source, Span<FixedLenByteArray> destination, decimal multiplier, ByteBuffer byteBuffer, int typeLength)
+        {
+            for (int i = 0; i < source.Length; ++i)
+            {
+                var byteArray = byteBuffer.Allocate(typeLength);
```
marcin-krystianc, line 486: Not sure if it really matters, but I think we can make it faster by allocating a single buffer for all values?

csharp/LogicalWrite.cs
```diff
@@ -449,6 +479,85 @@ public static void ConvertDecimal128(ReadOnlySpan<decimal?> source, Span<short>
...
+            for (int i = 0, dst = 0; i < source.Length; ++i)
+            {
+                var value = source[i];
+                if (value == null)
+                {
+                    defLevels[i] = nullLevel;
+                }
+                else
+                {
+                    var byteArray = byteBuffer.Allocate(typeLength);
```
marcin-krystianc, line 503: Here is the same.

marcin-krystianc (approved):

csharp/LogicalWrite.cs
adamreeve: The pointer to the byteArray gets stored in the FixedLengthByteArray values that are then read from the C++ library, so we couldn't reuse the same byteArray with length kept as typeLength, but I guess we could allocate a block of byteArray * source.Length which might be a bit faster by reducing the overhead of calls `byteBuffer.Allocate`. Is that what you mean?

We use the same pattern in a few other places, although usually via the `FromFixedLength` method, so I think this should be a separate PR if it does speed things up.

marcin-krystianc: > but I guess we could allocate a block of byteArray * source.Length which might be a bit faster by reducing the overhead of calls `byteBuffer.Allocate`. Is that what you mean?

Yes, exactly, this is what I meant. Possibly, the speed improvement is going to be minimal anyway. It is definitely not necessary to amend this PR, it is good as it is now.

PR #489, opened 2025-01-12
Fix CI breakage following `macos-12` runner image deprecation

jgiannuzzi:
The `macos-12` image we use for building has been [deprecated](https://github.com/actions/runner-images/issues/10721). This PR replaces its use by `macos-13` for building, and remove the use of `macos-12` for testing.

The `nudge` workflow configuration is also adjusted to notify us when the main CI workflow is cancelled, which was its conclusion when using a deprecated image, rather than a failure.

### DIFF
--- a/.github/workflows/nudge.yml
+++ b/.github/workflows/nudge.yml
@@ -15,3 +15,4 @@ jobs:
         uses: pavlovic-ivan/octo-nudge@v2
         with:
           webhooks: ${{ secrets.NUDGE_WEBHOOKS }}
+          conclusions: failure,cancelled
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
     if: github.event_name == 'schedule' || github.event_name == 'push' || github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id
     strategy:
       matrix:
-        os: [ubuntu-20.04, macos-12, windows-2022]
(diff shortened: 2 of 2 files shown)


marcin-krystianc (approved):
LGTM

.github/workflows/nudge.yml
```diff
@@ -15,3 +15,4 @@ jobs:
         uses: pavlovic-ivan/octo-nudge@v2
         with:
           webhooks: ${{ secrets.NUDGE_WEBHOOKS }}
+          conclusions: failure,cancelled
```
marcin-krystianc, line 18: Hi @pavlovic-ivan , it turns out that cancelled is not a valid conclusions value -> https://github.com/pavlovic-ivan/octo-nudge/blob/329060af1fb848e1ab0b3350dac3f5f5d9487eb9/util/index.js#L4. Is something that we could add to the nudge action or maybe it is not really possible to react on cancelled conclusions?

pavlovic-ivan: Hi Marcin. `cancelled` is not currently supported by the `octo-nudgde`. However, i can add it and support it

pavlovic-ivan: Hi @marcin-krystianc @jgiannuzzi . Please update the version of the octo-nudge action to `v3`. You should get the notifications for `cancelled` runs as well. Do let me know if it's not working (i've tested it, and it works in my setup).

PR #490, opened 2025-01-13
Cancelled is not a valid value for a conclusion in octo-nudge

marcin-krystianc:
As per the comment on https://github.com/G-Research/ParquetSharp/pull/489/files#r1912894991, 'cancelled'  value for conclusions is not supported by octo-nudge.


jgiannuzzi:
Sorry about that, I should have tested it!

PR #493, opened 2025-01-14
Workaround for breaking vcpkg changes

marcin-krystianc:
There is an ongoing process for reorganizing vcpkg, but it broke our CI:
- https://github.com/microsoft/vcpkg/commit/358f3893843c9d0d1b2e594183481fb68e3aa963
- https://github.com/microsoft/vcpkg-tool/pull/1553

I'm unsure how to fix it properly, but with this change, we at least make sure that we use vcpkg repo prior to the `vcpkgTools.xml` related changes and unblock our CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,8 +66,10 @@ jobs:
 
     # Method 3: Using printenv
     - name: Display All Environment Variables
-      run: printenv | sort
       shell: bash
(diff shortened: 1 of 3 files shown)


adamreeve (approved):
This looks fine as a temporary workaround but we should figure out how to make our CI work with the latest vcpkg soon

PR #496, opened 2025-01-20
Improved workaround for breaking vcpkg changes

adamreeve:
Rather than checking out an older version of vcpkg, we can run bootstrap-vcpkg to update the vcpkg tools.


marcin-krystianc (approved):
Nice 👍

PR #498, opened 2025-01-22
Set up API Reference website with DocFX

jescalada:
This PR sets up DocFX and related configs, such as Table of Contents files (`toc.yml`). It also sets up a GitHub Action that uploads the docs to the repo's URL (in this case, https://g-research.github.io/ParquetSharp). This may require the repo's settings (Settings > Pages > Build and Deployment) so that it deploys from GitHub Actions.

Right now, we can see a preview of the generated site from my own branch: https://jescalada.github.io/ParquetSharp/index.html

Note that the homepage is not yet set up. Should I set this to the main README for the time being? We could otherwise write up a more succinct introduction for ParquetSharp.

I also added a "Getting Started" file to the docs (because DocFX apparently doesn't read the main README). I simply copied the Quickstart section for this.

Feel free to suggest changes to how the page is structured or what to add to the "landing page".


adamreeve:
Nice! I haven't had a thorough look through everything but have some initial comments:
* I don't think we want the full README to be copied to the index page, but could maybe copy the "Introduction", "Why Use Parquet?" and "Quick Start" sections (rather than have a separate quick start page). And as your comment suggests, flesh this out with some more detailed installation and basic usage instructions but that could be added later.
* There are links at the bottom of the "Getting Started" page to the various documentation pages but they're broken. I think it would be nice to fix these and keep them on the index page.
* Can we change the logo in the header? (There are some in https://github.com/G-Research/ParquetSharp/tree/master/logo/svg)
* It looks like we should be able to use cross-references to autogenerated class docs but it wasn't clear to me exactly how that works (https://dotnet.github.io/docfx/docs/links-and-cross-references.html). Can you see if you can add a link to the `ParquetSharp.Arrow.FileReader` docs at [this line](https://github.com/G-Research/ParquetSharp/blob/6a0bea7391f2e7026f3bd80ad57d76d2acbcbed0/docs/Arrow.md?plain=1#L12C46-L12C75) for example?

jescalada:
I looked into implementing the auto-linking for the docs. Unfortunately, it works only with classes (not methods), forcing you to use the full qualifier for the class. I modified the references in the docs, and made a script so that the links are generated in case we have any references to `ParquetSharp` or `System`.

There is an additional caveat to this: something like "We will instantiate *an* `EncryptionConfiguration` (...)" would be shown as "We will instantiate **an** `ParquetSharp.Encryption.EncryptionConfiguration` (...)" in the Markdown. That said, the website will display the grammatically correct version without the full name (and link it to the docs).

Let me know if we'll be banking on users using the website docs rather than the GitHub docs, which will look a bit inconsistent with the new full-name linking method.

adamreeve:
> Let me know if we'll be banking on users using the website docs rather than the GitHub docs, which will look a bit inconsistent with the new full-name linking method.

I think it's fine as is, we'd probably try to direct people to the website rather than viewing the docs through GitHub.

adamreeve (approved):

PR #499, opened 2025-01-24
SchemaUtils Nested Map Fix

jescalada:
This PR fixes the issue where files generated by 3rd-party tools (with nonstandard schemas) fail to read correctly.

Let me know if the new test is comprehensive enough. I tried to add another test by generating a nested schema recursively, but this would require modifying `DoRoundtripTest` by changing the keys/values from 2D to 3D arrays.

This was the schema generation function I thought of, but I don't know if this is overkill for our purposes.

```csharp
        private static GroupNode CreateLogicalTypeMapSchema(bool nested)
        {
            using var stringType = LogicalType.String();
            using var mapType = LogicalType.Map();

            using var keyNode = new PrimitiveNode("key", Repetition.Required, stringType, PhysicalType.ByteArray);
            using var primitiveValueNode = new PrimitiveNode("value", Repetition.Optional, stringType, PhysicalType.ByteArray);
            using var valueNode = nested
                ? new GroupNode("value", Repetition.Optional, new Node[] {keyNode, primitiveValueNode}, mapType)
                : (Node)primitiveValueNode;
            using var keyValueNode = new GroupNode(
                "key_value", Repetition.Repeated, new Node[] {keyNode, valueNode}, mapType);
            using var colNode = new GroupNode(
                "col1", Repetition.Required, new Node[] {keyValueNode}, mapType);

            return new GroupNode(
                "schema", Repetition.Required, new Node[] {colNode});
        }
```


csharp.test/TestMaps.cs
```diff
@@ -49,6 +49,15 @@ public static void CanRoundtripRequiredMaps()
             DoRoundtripTest(false, keys, values);
         }
 
+        [Test]
+        public static void CanRoundtripLogicalMapType()
```
adamreeve, line 53: The logical map type is used in other test cases too but only on the top-level schema node, so this name is a little confusing. Maybe something like:
```suggestion
        public static void CanRoundtripNonStandardMapAnnotation()
```

csharp.test/TestMaps.cs
```diff
@@ -49,6 +49,15 @@ public static void CanRoundtripRequiredMaps()
...
+        {
+            var keys = new[] {new[] {"k1", "k2"}, new[] {"k3", "k4"}, Array.Empty<string>()};
+            var values = new[] {new[] {"v1", "v2"}, new[] {"v3", "v4"}, Array.Empty<string>()};
+
+            DoRoundtripTest(false, keys, values, CreateMapSchema(false));
```
adamreeve, line 58:
```suggestion
            DoRoundtripTest(false, keys, values, CreateMapSchema(true));
```

csharp.test/TestMaps.cs
```diff
@@ -220,15 +229,15 @@ private static void DoRoundtripTest(bool optional, string[][] keys, string[][] v
             Assert.AreEqual(0, pool.BytesAllocated);
         }
 
-        private static GroupNode CreateMapSchema(bool optional)
+        private static GroupNode CreateMapSchema(bool optional, bool useLogicalMapType = false)
```
adamreeve, line 232:
```suggestion
        private static GroupNode CreateMapSchema(bool optional, bool extraLogicalMapType = false)
```

adamreeve (changes requested):
This looks good thanks Juan, I just have some comments about the test

csharp.test/TestMaps.cs
jescalada: Oh, that was a little accident when I decided to refactor the existing `CreateMapSchema` instead of writing a new one. It's supposed to be `CreateMapSchema(false, true)`, where the second parameter is the new `nested` option.

adamreeve: Oh right, sorry I misread that. Looks good now

csharp.test/TestMaps.cs
```diff
@@ -190,7 +199,7 @@ private static void DoRoundtripTest(bool optional, string[][] keys, string[][] v
                 {
                     using var propertiesBuilder = new WriterPropertiesBuilder();
                     using var writerProperties = propertiesBuilder.Build();
-                    using var schemaNode = CreateMapSchema(optional);
+                    schemaNode ??= CreateMapSchema(optional);
```
adamreeve, line 202: This is getting quite nitpicky, but it might be cleaner to always require the schema node to be passed in, then the `optional` parameter can be removed from `DoRoundtripTest`, and rather than creating the map schema inline in the call to `DoRoundtripTest` we can create it with a `using` statement to make sure the native handle is always freed immediately rather than requiring the finalizer to be called.

adamreeve (approved):

PR #509, opened 2025-03-13
Add support for setting field_id on nodes

CurtHagenlocher:
There's currently no way to set the field_id when creating a new node. Ability to do this is required under some circumstances when writing Delta tables.

### DIFF
--- a/csharp/PublicAPI/net471/PublicAPI.Shipped.txt
+++ b/csharp/PublicAPI/net471/PublicAPI.Shipped.txt
@@ -678,7 +678,8 @@ ParquetSharp.Schema.GroupNode.FieldCount.get -> int
 ParquetSharp.Schema.GroupNode.FieldIndex(ParquetSharp.Schema.Node! node) -> int
 ParquetSharp.Schema.GroupNode.FieldIndex(string! name) -> int
 ParquetSharp.Schema.GroupNode.Fields.get -> ParquetSharp.Schema.Node![]!
-ParquetSharp.Schema.GroupNode.GroupNode(string! name, ParquetSharp.Repetition repetition, System.Collections.Generic.IReadOnlyList<ParquetSharp.Schema.Node!>! fields, ParquetSharp.LogicalType? logicalType = null) -> void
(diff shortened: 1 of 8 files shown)


CurtHagenlocher:
I didn't add field_id to the RowOriented API because it seems like a more advanced feature; let me know if I should do that too.

csharp/PublicAPI/net471/PublicAPI.Shipped.txt
```diff
@@ -678,7 +678,8 @@ ParquetSharp.Schema.GroupNode.FieldCount.get -> int
 ParquetSharp.Schema.GroupNode.FieldIndex(ParquetSharp.Schema.Node! node) -> int
 ParquetSharp.Schema.GroupNode.FieldIndex(string! name) -> int
 ParquetSharp.Schema.GroupNode.Fields.get -> ParquetSharp.Schema.Node![]!
-ParquetSharp.Schema.GroupNode.GroupNode(string! name, ParquetSharp.Repetition repetition, System.Collections.Generic.IReadOnlyList<ParquetSharp.Schema.Node!>! fields, ParquetSharp.LogicalType? logicalType = null) -> void
+ParquetSharp.Schema.GroupNode.GroupNode(string! name, ParquetSharp.Repetition repetition, System.Collections.Generic.IReadOnlyList<ParquetSharp.Schema.Node!>! fields, ParquetSharp.LogicalType? logicalType) -> void
+ParquetSharp.Schema.GroupNode.GroupNode(string! name, ParquetSharp.Repetition repetition, System.Collections.Generic.IReadOnlyList<ParquetSharp.Schema.Node!>! fields, ParquetSharp.LogicalType? logicalType = null, int fieldId = -1) -> void
```
adamreeve, line 682: :+1: I was initially worried this was a breaking change, but it looks like this should maintain source and binary compatibility as the default value is part of the source API but not the ABI. For my own reference, there are docs on adding optional parameters at https://github.com/dotnet/roslyn/blob/main/docs/Adding%20Optional%20Parameters%20in%20Public%20API.md

adamreeve (approved):
:+1: thanks Curt, this looks good to me.

I agree this doesn't need to be added to the row oriented API or the `Column` type as it's a more advanced feature. And we could always add it later if there's demand.

PR #516, opened 2025-04-10
Enable Control Flow Guard for windows builds

johnthcall:
Enable control flow guard when building with msvc as ParquetSharpNative.dll is reporting [BA2008](https://github.com/microsoft/binskim/blob/main/docs/BinSkimRules.md#rule-BA2008EnableControlFlowGuard) from BinSkim.

Verified that after this change that binskim no longer reports BA2008.

More details for /guard command can be found [here](https://learn.microsoft.com/en-us/cpp/build/reference/guard-enable-control-flow-guard?view=msvc-170).

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -104,6 +104,10 @@ add_definitions(-DARROW_STATIC)
 add_definitions(-DARROW_NO_DEPRECATED_API)
 add_definitions(-DPARQUET_STATIC)
 
+if (MSVC AND (MSVC_VERSION GREATER 1900))
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /guard:cf")
+endif()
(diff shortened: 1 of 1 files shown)


adamreeve:
Thanks @johnthcall, this seems like reasonable change. I'd like to first check the benchmarks on Windows before merging this.

adamreeve (approved):

adamreeve:
I didn't see any significant change in benchmarks with this change so have merged this now. Thanks @johnthcall

johnthcall:
Thanks @adamreeve Is there a cadence for package release or is it done ad-hoc?

adamreeve:
We don't really have a regular release cadence. We try to align with Arrow C++ releases but often end up skipping versions. The 20.0.0 version of Arrow is due to be released soon so I'll aim to get a release out soon after that.

PR #519, opened 2025-04-23
Add ARM64 Support to ParquetSharp

### LINKED ISSUE #441: [FEATURE REQUEST]: Windows ARM Support
### Is your feature request related to a problem? Please describe.

I'm trying to read parquet files with Power BI Desktop + Windows ARM (specifically through a Parallels VM with Apple Silicon), but ParquetSharp doesn't support Windows ARM.

### Describe the solution you'd like

Is Windows ARM support planned for ParquetSharp? There are new Qualcomm ARM Windows laptops being released in a couple months, so I imagine Windows ARM will grow in popularity.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

JoshuaOloton:
## Description
This PR adds support for building ParquetSharp on the ARM64 architecture. It includes modifications to the build system, dependency management, and resolves compatibility issues specific to ARM64 platforms.

## Checklist
- [x] Added new PowerShell build script specifically for ARM64 Windows
- [x] Updated `vcpkg.json` to include required dependencies for ARM64 builds.
- [x] Modified `CMakeLists.txt` to configure necessary dependencies for the ARM64 target.
- [x] Resolved all errors during the ARM64 build process.
- [x] Added build configuration to properly handle both Debug and Release builds on ARM64

## Related Issues
- ### Fixes #441

### DIFF
--- a/build_windows_arm.ps1
+++ b/build_windows_arm.ps1
@@ -1,0 +1,86 @@
+Set-StrictMode -Version 3
+$ErrorActionPreference = "Stop"
+
+# Find vcpkg or download it if required
+if ($null -ne $Env:VCPKG_INSTALLATION_ROOT) {
+  $vcpkgDir = $Env:VCPKG_INSTALLATION_ROOT
+  Write-Output "Using vcpkg at $vcpkgDir from VCPKG_INSTALLATION_ROOT"
+}
+elseif ($null -ne $Env:VCPKG_ROOT) {
+  $vcpkgDir = $Env:VCPKG_ROOT
+  Write-Output "Using vcpkg at $vcpkgDir from VCPKG_ROOT"
+}
+else {
+  $vcpkgDir = "$(Get-Location)/build/vcpkg"
+  Write-Output "Using local vcpkg at $vcpkgDir"
+  if (-not (Test-Path $vcpkgDir)) {
+    git clone https://github.com/microsoft/vcpkg.git $vcpkgDir
+    if (-not $?) { throw "git clone failed" }
+    & $vcpkgDir/bootstrap-vcpkg.bat
+    if (-not $?) { throw "bootstrap-vcpkg failed" }
+  }
+}
+
+$triplet = "arm64-windows"
+
+$build_types = @("Debug", "Release")
+
+if ($null -eq $env:ARROW_HOME) {
+    Write-Host "Error: ARROW_HOME environment variable not set." -ForegroundColor Red
+    Write-Host "Please set it to your Arrow installation directory, for example:" -ForegroundColor Yellow
+    Write-Host "    $env:ARROW_HOME = 'C:\path\to\arrow\install\$triplet'" -ForegroundColor Yellow
+    exit 1
+}
+
+$arrow_install_dir = $env:ARROW_HOME
+
+if (-not (Test-Path "$arrow_install_dir/release") -or -not (Test-Path "$arrow_install_dir/debug")) {
+    Write-Host "Error: ARROW_HOME directory doesn't contain expected 'debug' and 'release' subdirectories." -ForegroundColor Red
+    Write-Host "Please ensure your Arrow installation has both debug and release builds." -ForegroundColor Yellow
+    exit 1
+}
+
+Write-Host "Using Arrow installation from: $arrow_install_dir" -ForegroundColor Green
+
+$options = @()
+if ($Env:GITHUB_ACTIONS -eq "true") {
+  $build_types = @("Release")
+  $customTripletsDir = "$(Get-Location)/build/custom-triplets"
+  New-Item -Path $customTripletsDir -ItemType "directory" -Force > $null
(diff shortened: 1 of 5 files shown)


build_windows_arm.ps1
JoshuaOloton, line 1: A new Windows Powershell script was introduced to support ARM64 builds, primarily targeting arm64-windows triplets instead of the previous x64-windows-static triplet.
Some of the key changes introduces:
- The **build system** was switched from Visual Studio generator (-G "Visual Studio 17 2022" -A "x64") to Ninja Multi-Config generator (-G "Ninja Multi-Config"), for faster and more flexible builds with better support on Windows ARM64.
- **Compiler change**: Clang was chosen as the compiler (`CMAKE_C_COMPILER=clang-cl, CMAKE_CXX_COMPILER=clang-cl`).
- Since Arrow is now built from source, the script prompts users to set ARROW_HOME to the Arrow install directory, and ensures the debug/release sub-directories exist before continuing.

cpp/arrow/FileWriter.cpp
JoshuaOloton, line 1: This update was made to address the following warning:

```
[48/56] Building CXX object cpp\CMakeFiles\ParquetSharpNative.dir\arrow\FileWriter.cpp.obj
C:\Users\JoshuaOloton\Desktop\ParquetSharp\cpp\arrow\FileWriter.cpp(118,43): warning: 'NewRowGroup' is deprecated: Deprecated in 19.0.0. Use NewRowGroup() without the `chunk_size` argument. [-Wdeprecated-declarations]
  118 | 	TRYCATCH(PARQUET_THROW_NOT_OK(writer->NewRowGroup(chunk_size));)
  	|                                       	^
C:\Program Files (x86)\arrow\include\parquet/arrow/writer.h(95,3): note: 'NewRowGroup' has been explicitly marked deprecated here
   95 |   ARROW_DEPRECATED(
  	|   ^
C:\Program Files (x86)\arrow\include\arrow/util/macros.h(140,35): note: expanded from macro 'ARROW_DEPRECATED'
  140 | #  define ARROW_DEPRECATED(...) [[deprecated(__VA_ARGS__)]]
  	|                               	^
1 warning generated.
```

cpp/CMakeLists.txt
JoshuaOloton, line 1: I refactored the native CMake configuration to reference the locally built Arrow and Parquet libraries directly. Instead of using `find_package`, it now specifies the include directories and library paths for Debug and Release builds via the `ARROW_ROOT_DEBUG` and `ARROW_ROOT_RELEASE` variables. 
This was to ensure the project links against the correct Arrow build from source.

JoshuaOloton:
Hello @adamreeve 
I'm facing a blocker running the tests after building successfully on Windows ARM. Multiple testcases fail with the following error:

```
System.DllNotFoundException : Unable to load DLL 'ParquetSharpNatived' or one of its dependencies: The specified module could not be found. (0x8007007E)
Stack Trace (example):
   at ParquetSharp.WriterPropertiesBuilder.WriterPropertiesBuilder_Create(IntPtr& builder)
   at ParquetSharp.WriterPropertiesBuilder..ctor() in C:\Users\mlhuser\Desktop\ParquetSharp\csharp\WriterPropertiesBuilder.cs:line 18
   ...
```
I’ll keep digging into it, but please let me know if there’s anything obvious I might be missing in my build setup.

adamreeve:
Are you seeing those errors with .NET Framework? There's an older way to configure the DLL location used by .NET Framework that you might need to set up, see the `csharp/ParquetSharp.targets` file.

adamreeve:
Looking at this a bit more closely, I see that it relies on having a pre-built version of Arrow. Do you have a script you can share or add to this PR that shows how to build Arrow on arm64? 

This PR is a very helpful step towards getting full Windows arm64 support in ParquetSharp, but there are a few more steps before we can properly integrate this. We'll need to be able to build Arrow as well as the ParquetSharp native library in our CI process using the same Arrow C++ version as other platforms, and ideally need minimal work to maintain this process as new Arrow C++ versions are released.

One way to do this might be to use a custom vcpkg triplet that implements any customisation required to build Arrow on arm64 without needing to build it separate from vcpkg. Eg. you can use the `PORT MATCHES` condition in a custom triplet file to set options for a specific port. See https://learn.microsoft.com/en-us/vcpkg/users/triplets#per-port-customization

JoshuaOloton:
That seems like a much cleaner and maintainable approach long-term. 
Yes, it currently relies on an already built version of Arrow, I’ll share the script I used to build Arrow on Arm64 and its dependencies. I'll also look into implementing the custom vcpkg triplet as you suggested.

JoshuaOloton:
> Are you seeing those errors with .NET Framework? There's an older way to configure the DLL location used by .NET Framework that you might need to set up, see the `csharp/ParquetSharp.targets` file.

 Thanks! I think I may have spotted the issue — I’ll add a new `Content` block in `ParquetSharp.targets` specifically for Windows ARM under .NET Framework. I’ll report back once I’ve tested it.

PR #526, opened 2025-05-15
Support writing custom types that can represent nulls

### LINKED ISSUE #525: [FEATURE REQUEST]: Support custom "Nullable" types.
### Is your feature request related to a problem? Please describe.

We use a custom value type to represent dynamically typed values and I want to export these directly to a column in a Parquet file.

However, It seems that ParquetSharp will only allow logical types to be either reference types or `Nullable<>` instances for optional columns. 

Our values are Value Types, and can represent nulls directly.  We are using NaN-tagged doubles, but for sake of argument, assume a union type something like this:

```
enum TypeCode
{
    Null,
    Float,
    Int32,
}

[StructLayout(LayoutKind.Explicit)]
struct Value
{
    [FieldOffset(0)]
    public TypeCode Type;
    
    [FieldOffset(4)]
    public float FloatValue;

    [FieldOffset(4)]
    public int Int32Value;
}
```

I have implemented a `LogicalWriteConverterFactory` which would allow me to write these values directly to parquet data:

```
public class ValueLogicalWriteConverterFactory : LogicalWriteConverterFactory
{
    public override Delegate GetConverter<TLogical, TPhysical>(
        ColumnDescriptor columnDescriptor,
        ByteBuffer? byteBuffer)
    {
        if(typeof(TLogical) == typeof(Value))
        {
            if(columnDescriptor.PhysicalType == PhysicalType.Float )
                return (LogicalWrite<Value, float>.Converter)WriteFloat;
            
            // ...
        }

        return base.GetConverter<TLogical, TPhysical>(columnDescriptor, byteBuffer);
    }

    private static void WriteFloat(
        ReadOnlySpan<Value> source,
        Span<short> defLevels,
        Span<float> destination,
        short nullLevel)
    {
        int num = 0;
        for (int index = 0; index < source.Length; ++index)
        {
            var value = source[index];
            switch (value.Type)
            {
                case TypeCode.Null when !defLevels.IsEmpty:
                    defLevels[index] = nullLevel;
                    break;
                    
                case TypeCode.Float:
                    destination[num++] = value.FloatValue;
                    defLevels[index] = (short)(nullLevel + 1);
                    break;
                
                default:
                    throw new InvalidOperationException();
            }
        }
    }
}
```

And then I was hoping that I could write these values directly to a column:

```
var data = new[] { new Value(1), default(Value), new Value(2), new Value(3) };

var columns = new[] { new PrimitiveNode("column", Repetition.Optional, LogicalType.None(), PhysicalType.Float) };
var schema = new GroupNode("schema", Repetition.Required, columns);

using var writer = new ParquetFileWriter("out.parquet", schema,WriterProperties.GetDefaultWriterProperties());
writer.LogicalWriteConverterFactory = new ValueLogicalWriteConverterFactory();

using var rowGroup = writer.AppendRowGroup();
using var column = rowGroup.NextColumn();
using var logicalWriter = column.LogicalWriterOverride<Value>();

logicalWriter.WriteBatch(data);

writer.Close();
```

However, when you run this you get an exception from `LogicalWriterOverride<Value>()`:

```
Failed to create a batch writer for type System.Nullable`1[Value]
```

### Describe the solution you'd like

Ideally, the restriction should be lifted and the call to `column.LogicalWriterOverride<Value>()` should succeed. 

### Describe alternatives you've considered

The obvious workaround is that we can copy batches of values from our `Value[]` instances into an intermediate buffer (e.g. of type `Value?[]` or `float?[]`) before passing them to the logical writer to serialize.  This does have a performance hit, but is admittedly fine for our purposes,.

However, given the goal of ParquetSharp to be as performant as possible, and given that the ValueLogicalWriteConverterFactory presumably exists to support these kinds of use cases, it's a shame that it falls over at this last hurdle.

### Additional context

_No response_

adamreeve:
Fixes #525

I'm not 100% sure about this approach and whether we want to catch any discrepancies between the element type returned by `GetSystemTypes` and what the user has provided as an override. Eg. what should happen if the schema represents a list but the element type override doesn't use an array?

I also tried an alternative approach in 46eddf24ee34261b2c993c39a1fbea3645525565 that would require the user to specify whether a type can represent nulls in the `LogicalTypeFactory`.

### DIFF
--- a/csharp.test/TestCustomNullableType.cs
+++ b/csharp.test/TestCustomNullableType.cs
@@ -27,7 +27,6 @@ namespace ParquetSharp.Test
             {
                 using var writer = new ParquetFileWriter(outStream, schema, WriterProperties.GetDefaultWriterProperties());
                 writer.LogicalWriteConverterFactory = new ValueLogicalWriteConverterFactory();
-                writer.LogicalTypeFactory = new ValueLogicalTypeFactory();
 
                 using var rowGroup = writer.AppendRowGroup();
                 using var column = rowGroup.NextColumn();
@@ -41,7 +40,6 @@ namespace ParquetSharp.Test
             using var input = new BufferReader(buffer);
             using var fileReader = new ParquetFileReader(input);
             fileReader.LogicalReadConverterFactory = new ValueLogicalReadConverterFactory();
-            fileReader.LogicalTypeFactory = new ValueLogicalTypeFactory();
 
             using var groupReader = fileReader.RowGroup(0);
             using var columnReader = groupReader.Column(0).LogicalReaderOverride<Value>();
@@ -138,19 +136,6 @@ namespace ParquetSharp.Test
                 {
                     destination[i] = defLevels[i] != definedLevel ? default : new Value(source[src++]);
                 }
-            }
-        }
-
-        private class ValueLogicalTypeFactory : LogicalTypeFactory
-        {
-            public override bool IsNullable(Type type)
-            {
-                if (type == typeof(Value))
-                {
-                    return true;
-                }
-
-                return base.IsNullable(type);
             }
         }
     }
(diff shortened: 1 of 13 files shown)


marcin-krystianc (approved):
LGTM 👍

PR #528, opened 2025-05-20
Upgrade Arrow C++ to 20.0.0

adamreeve:


csharp.test/TestWriterProperties.cs
```diff
@@ -41,8 +41,8 @@ public static void TestPropertiesBuilder()
                 .MaxRowGroupLength(789)
                 .Version(ParquetVersion.PARQUET_1_0)
                 .WriteBatchSize(666)
-                .DisableWritePageIndex()
                 .EnableWritePageIndex()
+                .DisableWritePageIndex()
```
marcin-krystianc, line 45: Is this change necessary?

csharp.test/TestWriterProperties.cs
```diff
@@ -56,7 +56,7 @@ public static void TestPropertiesBuilder()
             Assert.AreEqual(789, p.MaxRowGroupLength);
             Assert.AreEqual(ParquetVersion.PARQUET_1_0, p.Version);
             Assert.AreEqual(666, p.WriteBatchSize);
-            Assert.True(p.WritePageIndex);
+            Assert.False(p.WritePageIndex);
```
marcin-krystianc, line 59: Is this change necessary?

csharp.test/TestWriterProperties.cs
```diff
@@ -112,7 +112,7 @@ public static void TestOverrideDefaults()
                 DefaultWriterProperties.MaxRowGroupLength = 789;
                 DefaultWriterProperties.Version = ParquetVersion.PARQUET_1_0;
                 DefaultWriterProperties.WriteBatchSize = 666;
-                DefaultWriterProperties.WritePageIndex = true;
+                DefaultWriterProperties.WritePageIndex = false;
```
marcin-krystianc, line 115: Is this change necessary?

csharp.test/TestWriterProperties.cs
```diff
@@ -130,7 +130,7 @@ public static void TestOverrideDefaults()
                 Assert.AreEqual(789, p.MaxRowGroupLength);
                 Assert.AreEqual(ParquetVersion.PARQUET_1_0, p.Version);
                 Assert.AreEqual(666, p.WriteBatchSize);
-                Assert.True(p.WritePageIndex);
+                Assert.False(p.WritePageIndex);
```
marcin-krystianc, line 133: Is this change necessary?

marcin-krystianc (approved):
I've spot a few seemingly unnecessary changes, looks good otherwise.

csharp.test/TestWriterProperties.cs
adamreeve: This was flipping the order of operations so that the end state is different to the new default, to check that these methods have an effect.

csharp.test/TestWriterProperties.cs
adamreeve: Same as above, this is changing the test to set the value to something different to the default, which changed from false to true.

Issue #517, opened 2025-04-15
[FEATURE REQUEST]: Add sorting_columns


alxu-avpt:
### Is your feature request related to a problem? Please describe.

Since #163, upstream arrow supports sorting_columns now https://github.com/apache/arrow/commit/da6dbd48607089d716505054176e345b704570c5.

### Describe the solution you'd like

Add a proxy function, similar to `builder.Encoding` etc.

### Additional context

It's possible for ParquetSharp to check the data sort, but I feel it's not necessary.

adamreeve:
Hi @alxu-avpt, yes we should add this to ParquetSharp. Is this something you could make a PR to add?

I agree it's not necessary to check the sort order, we should just document that it's up to the user to ensure their sort order is correct.

PR #536, opened 2025-06-18
Add sorting columns

iksarfo:
Issue https://github.com/G-Research/ParquetSharp/issues/517

## Overview

This PR adds support for sorting columns in the C# writer APIs and brings the published public-API lists back into alignment. In particular:

- Introduce `WriterPropertiesBuilder.SortingColumns(int[] columnIndices, bool[] isDescending, bool[] nullsFirst)` to configure sort order.
- Expose `WriterProperties.SortingColumns()` to retrieve sorting metadata.
- Add `WriterProperties.GetDefaultWriterProperties()` factory method.
- Surface the two new hooks `ParquetFile.OnReadExpressionCreated` and `ParquetFile.OnWriteExpressionCreated`.
- Update all PublicAPI.Shipped.txt files (root, net6, netstandard2.1) with the new members so builds treating API as code will pass.
- Populate the root `PublicAPI.Unshipped.txt` with the corresponding entries to ensure future-API tracking.

## Changes

- **csharp/WriterPropertiesBuilder.cs**  
  - Added `SortingColumns(...)` method with null/length checks and pinned-memory interop.
  - Wire up defaults from `DefaultWriterProperties.Sorting*` in `ApplyDefaults()`.
- **csharp/WriterProperties.cs**  
  - Exposed `GetDefaultWriterProperties()`.
  - Implemented `(int[] ColumnIndices, bool[] IsDescending, bool[] NullsFirst) SortingColumns()` with native interop and free.
- **csharp/DefaultWriterProperties.cs**  
  - Added three new arrays: `SortingColumnIndices`, `SortingColumnsDescending`, `SortingColumnsNullsFirst`.
- **cpp/WriterPropertiesBuilder.cpp** & **cpp/WriterProperties.cpp**  
  - Bridge `Builder::set_sorting_columns(...)` and `WriterProperties::sorting_columns()` respectively.
- **csharp/PublicAPI.Shipped.txt** (root, net6, netstandard2.1)  
  - Added declarations for the new members and factory methods.
- **csharp/PublicAPI.Unshipped.txt**  
  - Listed the “unshipped” API changes so tooling will include them in the next release.

## Tests

- **csharp.test/TestWriterProperties.cs**  
  - New unit tests for `SortingColumns()` round-trip.
  - Existing tests for default properties and builder coverage pass unchanged.

## Validation

- All existing tests continue to pass.
- `dotnet build` on net6, netstandard2.1, net471 targets succeeds with no API mismatches.
- Verified native interop memory safety with manual pin-free in tests.
- No breaking changes for existing users—sorting is opt-in via builder.

## Related

- Closes #\<issue-number\> (introduce sorting-columns support)
- Follows on from #\<previous-PR\> (public-API cleanup)

---

*Please review the public-API updates carefully before merging.*


adamreeve (changes requested):
Thanks for the PR @iksarfo! I've requested some changes. I think before merging this I'd also like to see support for reading the sorting columns from the `RowGroupMetaData`, and a test that writes a file with sorting columns and then reads them back from the row group.

cpp/RowGroupMetaData.cpp
```diff
@@ -32,4 +31,58 @@ extern "C"
...
+		}
+		catch (const std::exception& e)
+		{
+			return new ExceptionInfo(typeid(e).name(), e.what());
+		}
+	}
+
+	PARQUETSHARP_EXPORT void RowGroupMetaData_Sorting_Columns_Free(int32_t* column_indices, bool* descending, bool* nulls_first)
+	{
+		if (column_indices != nullptr)
```
marcin-krystianc, line 73: I think that the null checks are not needed, as deleting a null pointer has no effect.

cpp/WriterProperties.cpp
```diff
@@ -121,4 +121,58 @@ extern "C"
...
+		}
+		catch (const std::exception& e)
+		{
+			return new ExceptionInfo(typeid(e).name(), e.what());
+		}
+	}
+
+	PARQUETSHARP_EXPORT void WriterProperties_Sorting_Columns_Free(int32_t* column_indices, bool* descending, bool* nulls_first)
+	{
+		if (column_indices != nullptr)
```
marcin-krystianc, line 163: Same, deleting a null pointer has no effect.

csharp/RowGroupMetaData.cs
```diff
@@ -35,6 +35,68 @@ internal RowGroupMetaData(IntPtr handle)
...
+                var result = new WriterProperties.SortingColumn[numColumns];
+                for (var i = 0; i < numColumns; i++)
+                {
+                    result[i] = new WriterProperties.SortingColumn(columnIndices[i], isDescending[i], nullsFirst[i]);
+                }
+                return result;
+            }
+            finally
+            {
+                if (columnIndicesPtr != IntPtr.Zero || descendingPtr != IntPtr.Zero || nullsFirstPtr != IntPtr.Zero)
```
marcin-krystianc, line 93: I think that the null checks are not needed, as deleting a null pointer has no effect.

cpp/RowGroupMetaData.cpp
```diff
@@ -32,4 +31,47 @@ extern "C"
 	{
 		TRYCATCH(*total_byte_size = row_group_meta_data->total_byte_size();)
 	}
+
+	PARQUETSHARP_EXPORT ExceptionInfo* RowGroupMetaData_Sorting_Columns(const RowGroupMetaData* row_group_meta_data, int32_t** column_indices, bool** descending, bool** nulls_first, int* num_columns)
+	{
+		try
```
adamreeve, line 37: Can you please use the `TRYCATCH` macro here

cpp/WriterProperties.cpp
```diff
@@ -121,4 +121,47 @@ extern "C"
 	{
 		TRYCATCH(*max_statistics_size = (*writer_properties)->max_statistics_size(*path);)
 	}
+
+	PARQUETSHARP_EXPORT ExceptionInfo* WriterProperties_Sorting_Columns(const std::shared_ptr<WriterProperties>* writer_properties, int32_t** column_indices, bool** descending, bool** nulls_first, int* num_columns)
+	{
+		try
```
adamreeve, line 127: This should use the `TRYCATCH` macro too

csharp.test/TestWriterProperties.cs
```diff
@@ -247,5 +247,92 @@ public static void TestByteStreamSplitEncodingWithNulls()
...
+            {
+                var writerProperties = new WriterPropertiesBuilder()
+                    .SortingColumns(sortingColumns)
+                    .Build();
+
+                using var fileWriter = new ParquetFileWriter(output, columns, writerProperties);
+                using var groupWriter = fileWriter.AppendRowGroup();
+
+                using var idWriter = groupWriter.NextColumn().LogicalWriter<int>();
+                idWriter.WriteBatch(new[] { 3, 1, 2 });
```
adamreeve, line 308: I'm nitpicking, but it would be nice if the example was sorted according to the sorting column metadata :smile: 
```suggestion
                idWriter.WriteBatch(new[] { 1, 2, 3 });
```

csharp/WriterProperties.cs
```diff
@@ -9,6 +9,79 @@ namespace ParquetSharp
...
+        /// </summary>
+        public readonly struct SortingColumn : IEquatable<SortingColumn>
+        {
+            /// <summary>
+            /// Creates a new sorting column specification.
+            /// </summary>
+            /// <param name="columnIndex">The index of the column to sort by</param>
+            /// <param name="isDescending">Whether to sort in descending order (true) or ascending order (false)</param>
+            /// <param name="nullsFirst">Whether nulls should come first (true) or last (false)</param>
+            public SortingColumn(int columnIndex, bool isDescending, bool nullsFirst)
```
adamreeve, line 23: For convenience I think this could have some default values:
```suggestion
            public SortingColumn(int columnIndex, bool isDescending = false, bool nullsFirst = false)
```

adamreeve (commented):
Thanks for the updates, this is looking really good! I've just got a few minor follow up comments

adamreeve:
I see the format check is failing. You should be able to fix this with:
```
dotnet tool restore
```

csharp.test/TestWriterProperties.cs
adamreeve: Claude Sonnet 4 doesn't know what it's talking about. Setting the sorting columns doesn't do any sorting, it's up to the user to set the sort order correctly.

csharp/WriterPropertiesBuilder.cs
```diff
@@ -475,6 +476,63 @@ public WriterPropertiesBuilder DisablePageChecksum()
             return this;
         }
 
+        /// <summary>
+        /// Set the sorting columns used to sort data when writing to a Parquet file.
```
adamreeve, line 480:
```suggestion
        /// Set the sorting columns to describe how written data is ordered.
        ///
        /// Note that no sorting or validation is done automatically. It is your responsibility to ensure
        /// that data is written in the specified order. 
```

csharp/WriterProperties.cs
```diff
@@ -143,6 +204,61 @@ public ulong MaxStatisticsSize(ColumnPath path)
         /// </summary>
         public bool PageChecksumEnabled => ExceptionInfo.Return<bool>(Handle, WriterProperties_Page_Checksum_Enabled);
 
+        /// <summary>
+        /// Gets the columns by which the data is sorted when writing to the file.
```
adamreeve, line 208:
```suggestion
        /// Gets the columns by which the data is sorted.
```

csharp.test/TestWriterProperties.cs
adamreeve: This made me realise I should double check the documentation on the new methods, so I've made some suggestions to make this clearer to uses.

cpp/WriterProperties.cpp
adamreeve: This was resolved but not done?

csharp/ParquetSharp.csproj
```diff
@@ -75,6 +75,7 @@
 
   <ItemGroup>
     <PackageReference Include="Apache.Arrow" Version="13.0.0" />
+    <PackageReference Include="Microsoft.Bcl.HashCode" Version="1.1.1" />
```
adamreeve, line 78: Can you move this down inside the conditional block (`<ItemGroup Condition="'$(TargetFramework)' == 'net471' or '$(TargetFramework)' == 'netstandard2.0'">`)

adamreeve (changes requested):

adamreeve:
Thanks for the latest changes @iksarfo. This is pretty close but there are still a few changes needed. I think the formatting also needs to be fixed as I mentioned above too

adamreeve (commented):
This looks great, thanks for the contribution @iksarfo! There's a minor merge conflict that needs fixing and the PublicAPI needs updating again but otherwise it should be good to go.

cpp/WriterPropertiesBuilder.cpp
```diff
@@ -209,4 +209,21 @@ extern "C"
...
+		std::vector<parquet::SortingColumn> sorting_columns;
+		
+		// Create SortingColumn objects for each column
+		for (int i = 0; i < num_columns; ++i)
+		{
+			parquet::SortingColumn column;
+			column.column_idx = column_indices[i];
+			column.descending = descending[i];
+			column.nulls_first = nulls_first[i];
+			sorting_columns.push_back(column);
```
marcin-krystianc, line 224: To avoid copying the stack variable, we can use:
```suggestion
			sorting_columns.emplace_back(column);
```

cpp/WriterPropertiesBuilder.cpp
```diff
@@ -209,4 +209,21 @@ extern "C"
 	{
 		TRYCATCH(builder->disable_page_checksum();)
 	}
+
+	PARQUETSHARP_EXPORT ExceptionInfo* WriterPropertiesBuilder_Sorting_Columns(WriterProperties::Builder* builder, int32_t* column_indices, bool* descending, bool* nulls_first, int num_columns)
+	{
+		std::vector<parquet::SortingColumn> sorting_columns;
+
```
marcin-krystianc, line 216: We know exactly how many elements there will be in the vector, so we can reserve necessary storage to avoid reallocations:
```suggestion
		sorting_columns.reserve(num_columns);
```

marcin-krystianc (approved):
Just left a minor comments about perf improvements.

cpp/WriterPropertiesBuilder.cpp
marcin-krystianc: > I'm not a C++ expert, but my understanding is using emplace_back like this doesn't help as we've already created the column, which will need to be copied.

Yeah, you are right. my suggestion wouldn't change anything.

marcin-krystianc (approved):

PR #537, opened 2025-06-26
Use ArrayPool for Reader and Writer Buffers

johnthcall:
In three places in the code base a buffer array is allocated when instead an array pool can be used instead to avoid large buffer allocation for each read/write.


marcin-krystianc (approved):
LGTM

adamreeve:
Thanks @johnthcall!

PR #540, opened 2025-07-07
Build Arrow with support for mimalloc and jemalloc

adamreeve:
* Enables building Arrow with mimalloc on all x64 platforms and Linux arm64
* Enables building Arrow with jemalloc on Linux and x64 macOS

This means the default allocator will now be mimalloc, matching upstream Arrow on platforms where we enable it. Users can switch allocators with the `ARROW_DEFAULT_MEMORY_POOL` environment variable.

Mimalloc and jemalloc should work on arm64, but it appears that these don't build correctly when cross-compiling on an x64 host as part of the Arrow build. Arrow doesn't use mimalloc or jemalloc from vcpkg but always builds its own version, so this probably needs some fixes to the upstream Arrow build scripts.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -7,7 +7,10 @@
     {
       "name": "arrow",
       "features": [
-        "mimalloc",
+        {
+          "name": "mimalloc",
+          "platform": "linux | !arm64"
+        },
         {
           "name": "jemalloc",
           "platform": "!windows&!arm64"
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -6,7 +6,13 @@
   "dependencies": [
     {
       "name": "arrow",
-      "features": ["jemalloc", "mimalloc"]
+      "features": [
+        "mimalloc",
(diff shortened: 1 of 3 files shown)


vcpkg.json
```diff
@@ -2,14 +2,26 @@
   "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
   "name": "parquetsharp",
   "version-string": "undefined",
-  "builtin-baseline": "d9ccd77bb554e67137f3f754a2e2f11f4188c82c",
+  "builtin-baseline": "e08b7bd89ae162f8579df2f8d39a1ae94107c8fd",
```
adamreeve, line 5: Needed to bump the baseline to pick up this fix that allows using mimalloc on Windows: https://github.com/microsoft/vcpkg/pull/46299

marcin-krystianc (approved):

PR #561, opened 2025-09-24
Run benchmarks in check mode in CI

adamreeve:


adamreeve:
I think the benchmarks are probably being killed due to using too much memory. Max RSS running them on my machine is > 23 GiB. It would be good to have a flag to reduce the size of data used to allow checking the benchmarks work in CI.

PR #568, opened 2025-10-03
ci: macos-13 runners are deprecated

jgiannuzzi:
- use the minimum non-deprecated macOS runner to build
- avoid cross-compilation because arrow-vendored
  jemalloc/mimalloc builds currently use the host architecture instead
  of the target - this needs to be fixed upstream in arrow
- enable jemalloc/mimalloc for arm64 architectures
- test against all stable free runner images

note: `macos-15-intel` is used to build macOS/x64, but the build is still compatible with older versions, as seen in [this run where tests were executed on `macos-13`](https://github.com/G-Research/ParquetSharp/actions/runs/18278992855/job/52039859098).


marcin-krystianc (approved):
LGTM.

PR #569, opened 2025-10-06
Add binary_type to ArrowReaderProperties

haianhng31:


csharp.test/Arrow/TestArrowReaderProperties.cs
```diff
@@ -28,12 +28,14 @@ public void TestSetProperties()
             properties.SetReadDictionary(0, true);
             properties.PreBuffer = false;
             properties.CoerceInt96TimestampUnit = Apache.Arrow.Types.TimeUnit.Microsecond;
+            properties.SetBinaryType(Apache.Arrow.Types.ArrowTypeId.Binary);
```
haianhng31, line 31: `Apache.Arrow.Types.ArrowTypeId` doesn't have `BinaryView` & `LargeBinary` so i cannot use them for testing here

PR #571, opened 2025-10-08
Update Apache.Arrow version to 18.0.0

haianhng31:


adamreeve (approved):
:+1: thanks @haianhng31

zhuangState:
I have to revert to 13.0.0, otherwise using 23.0.0.2 will cause an exception

System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime.CompilerServices.Unsafe, Version=4.0.4.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the specified file.

PR #581, opened 2025-10-19
Add set_max_statistics_size to WriterProperties

haianhng31:


csharp/WriterProperties.cs
```diff
@@ -335,7 +335,7 @@ public SortingColumn[] SortingColumns()
         private static extern IntPtr WriterProperties_Statistics_Enabled(IntPtr writerProperties, IntPtr path, [MarshalAs(UnmanagedType.I1)] out bool enabled);
 
         [DllImport(ParquetDll.Name)]
-        private static extern IntPtr WriterProperties_Max_Statistics_Size(IntPtr writerProperties, IntPtr path, [MarshalAs(UnmanagedType.I1)] out ulong maxStatisticsSize);
+        private static extern IntPtr WriterProperties_Max_Statistics_Size(IntPtr writerProperties, IntPtr path, [MarshalAs(UnmanagedType.U8)] out ulong maxStatisticsSize);
```
adamreeve: The `MarshalAs` attribute shouldn't be necessary as `U8` should be the default for a `ulong`. It looks like having `I1` previously was a copy-paste error.

adamreeve (approved):

PR #583, opened 2025-10-21
Dramatically faster CodeQL

jgiannuzzi:
- avoid huge caching artefact (1.5GiB per run!)
- in line with the latest CodeQL guidelines, do not build source code anymore
- this results in a 3min CodeQL run instead of 40-50min!

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -22,41 +22,12 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v5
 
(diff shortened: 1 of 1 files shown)


adamreeve (approved):
:+1: Nice.

It sounds like setting `build-mode` [shouldn't work with multiple languages](https://github.com/github/codeql-action/blob/0c5185d061ea8cb82b9c5cd84d2c51a7fc2d9d6b/init/action.yml#L34-L35), but it looks like it works just fine. And I couldn't see any docs about C# support for `build-mode: none`, but it looks like that works fine too.

PR #584, opened 2025-10-21
Enable vcpkg cache via vcpkg-cache-action

### LINKED ISSUE #562: Fix vcpkg caching in CI
See https://github.com/G-Research/ParquetSharp/actions/runs/17726976872/job/50369637414#step:14:118

The `x-gha` binary caching backend has been removed. GitHub packages based caching seems to be the recommended approach now: https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages

jgiannuzzi:
- a new GitHub Action is introduced to provide vcpkg binary caching via a custom
  HTTP server that proxies to GitHub Actions cache
- the result is granular caching of `vcpkg` artifacts, dramatically reducing build
  times
- the action is a typescript shim that invokes a Python script (this is so we
  can register a `post` step)
- the shim is "built" via `npm build` and the resulting JavaScript is checked
  into source control (best practice for GitHub Actions)
- the Python script is launched via `pipx` and `uv` (as `uv` itself is not installed
  on the GitHub Actions runners)
- `uv` is invoked with the `--locked` option to ensure consistent dependency
  resolution
- the bulk of the action is written in Python for cross-platform compatibility
  and fast startup times without requiring a separate compilation step
- the Python script is responsible for both the `main` and `post` steps of the
  action and sets the `VCPKG_BINARY_SOURCES` environment variable accordingly

fixes #562


.github/actions/vcpkg-cache-action/main.py
```diff
@@ -0,0 +1,197 @@
...
+    ]
+
+    log_fd, log_path = tempfile.mkstemp()
+    tail_task = asyncio.create_task(tail(log_path))
+    proc: subprocess.Popen | None = None
+
+    try:
+        # asyncio.subprocess does not allow us to create detached processes on
+        # Windows so we use subprocess.Popen here
+        proc = subprocess.Popen(
```
adamreeve: Should `creationflags=subprocess.DETACHED_PROCESS` be used on Windows? I guess it obviously works as-is so it's not needed, but maybe that would solve the problem with asyncio.subprocess?

.github/actions/vcpkg-cache-action/main.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+class ProxyStartError(RuntimeError):
+    pass
+
+
+def find_free_port() -> int:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind((LOCALHOST, 0))
+        return s.getsockname()[1]
```
adamreeve: I guess there's a very small chance of a race condition here if anything else was to try to bind to a port after this disconnects, compared to the approach of letting uvicorn pick a port and reading it from the log. I think this is fine though and a lot simpler.

adamreeve (approved):
This looks great thanks Jonathan. It took me a little while to make sense of everything but it looks like this works really nicely and the builds are so much faster!

Is there any way to test this locally? I guess you'd need something that acts like the GitHub actions API, so this would be complicated.

Some notes I made for myself on terms I wasn't familiar with:
* uvicorn - an ASGI web server that's pulled in as a dependency of FastAPI
* twirp - seems to refer to the [twirp RPC framework](https://github.com/twitchtv/twirp), which GitHub must be using for the cache API

.github/actions/vcpkg-cache-action/main.py
jgiannuzzi: I did try that with `asyncio.subprocess`, but it didn't work. I even tried adding more flags, like `CREATE_NEW_PROCESS_GROUP` and `CREATE_BREAKAWAY_FROM_JOB`, but it didn't work either. The issue seems to be with how the asyncio event loop closes the child process on exit, but only on Windows. I am not entirely sure!

marcin-krystianc (dismissed):

.github/actions/vcpkg-cache-action/main.py
jgiannuzzi: So the loop was already a `ProactorEventLoop` on Windows (otherwise launching a subprocess would not work at all), and in my testing it looks like it is indeed the one closing the child process, no matter which creation flags I pass. I'm giving up 🤷

PR #589, opened 2025-11-02
Add cache_options

haianhng31:


csharp/Arrow/ArrowReaderProperties.cs
```diff
@@ -225,6 +252,12 @@ public bool ArrowExtensionEnabled
         [DllImport(ParquetDll.Name)]
         private static extern IntPtr ArrowReaderProperties_SetArrowExtensionEnabled(IntPtr readerProperties, bool extensionsEnabled);
 
+        [DllImport(ParquetDll.Name)]
+        private static extern IntPtr ArrowReaderProperties_GetCacheOptions(IntPtr readerProperties, out ParquetSharp.CacheOption cacheOptions);
```
adamreeve: I'm a bit surprised this works without setting a `[StructLayout(LayoutKind.Sequential)]` attribute on the `CacheOption` type. In general, we probably can't rely on the layout used by .NET matching the layout used by C++ so this isn't safe. I think it would probably be simplest to do something similar to the `SetCacheOptions` method and get the various members of the options separately.

csharp/CacheOption.cs
```diff
@@ -0,0 +1,48 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace ParquetSharp
+{
+    public struct CacheOption
```
adamreeve: This should be called `CacheOptions` to match the C++ library and because it contains multiple options.

PR #593, opened 2025-11-28
Bump Arrow port version for compatibility with vcpkg master

adamreeve:
Looks like [setting `CMAKE_SYSTEM_PROCESSOR`](https://github.com/microsoft/vcpkg/pull/47958/files#diff-363df8fbbe7343b382e186887095c573c6a0e78120eba6cf6c1941747f0898f7L68) in the older port version caused builds to fail in CI due to https://github.com/microsoft/vcpkg/issues/48496 even though we aren't cross compiling: https://github.com/G-Research/ParquetSharp/actions/runs/19750737197/job/56593190898#step:12:5078

### DIFF
--- a/build_windows.ps1
+++ b/build_windows.ps1
@@ -59,7 +59,6 @@ if ($Env:GITHUB_ACTIONS -eq "true") {
   $options += "VCPKG_OVERLAY_TRIPLETS=$customTripletsDir"
 }
 
-
 $options += " -DCMAKE_VERBOSE_MAKEFILE=ON"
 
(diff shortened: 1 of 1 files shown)


adamreeve:
Bumping the port version fixed the Linux native library builds, but the macos-15-intel tests are now failling. There was a runner image update (https://github.com/actions/runner-images/releases/tag/macos-15%2F2025.1126), which includes an XCode update. And it looks like `catch (const std::exception& exception)` no longer catches a `std::out_of_range` exception???

adamreeve:
Closing, this has been superseded by #599

PR #594, opened 2025-11-30
Guide on optimizing Parquet file reading to reduce memory usage 

haianhng31:


docs/guides/ReadingMemoryOptimization.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+### Buffered Stream
+
+```csharp
+public static void ReadWithCustomBuffer(string filePath, int bufferSize = 1024 * 1024)
+{
+    // Use a buffered stream with custom buffer size (1 MB in this example)
+    using var fileStream = File.OpenRead(filePath);
+    using var bufferedStream = new BufferedStream(fileStream, bufferSize);
```
adamreeve: By using a buffered stream, I actually meant enabling it in the ReaderProperties with `ReaderProperties.EnableBufferedStream`.

In previous investigations I've found that this can significantly reduce memory usage.

I don't think using a .NET System.IO.BufferedStream will change memory usage characteristics much.

PR #630, opened 2026-03-26
Write benchmarks

larrytamnjong:


adamreeve (commented):
Thanks Larry, I've left a couple of suggestions

larrytamnjong:
Hi @adamreeve , I ran the tests with the three files and drafted some documentation of the findings. Please let me know what you think.

csharp.config.benchmarks/ParquetSharpConfigBenchmarks.cs
```diff
@@ -211,5 +211,80 @@ public static void RowOriented_Default()
...
+                Compression.Snappy => "Snappy",
+                Compression.Zstd => "Zstd",
+                _ => compression.ToString()
+            };
+            string outputFile = $"{baseName}_{encodingTag}_{compressionTag}.parquet";
+
+            byte[] rawBytes = File.ReadAllBytes(binPath);
+            int floatCount = rawBytes.Length / sizeof(float);
+            float[] values = new float[floatCount];
+            Buffer.BlockCopy(rawBytes, 0, values, 0, floatCount * sizeof(float));
```
adamreeve: We can avoid storing two copies of the data by reinterpreting the bytes instead of copying them:

```suggestion
            var values = MemoryMarshal.Cast<byte, float>(rawBytes.AsSpan());
```

csharp.config.benchmarks/ParquetSharpConfigBenchmarks.cs
```diff
@@ -211,5 +211,80 @@ public static void RowOriented_Default()
...
+
+            Console.WriteLine($"Read {values.Length:N0} floats from {binPath}");
+
+            var columns = new Column[]
+            {
+                new Column<int>("RowIndex"),
```
adamreeve: I don't think we should include this row index column, we want to focus just on the float data.

csharp.config.benchmarks/ParquetSharpConfigBenchmarks.cs
```diff
@@ -211,5 +211,80 @@ public static void RowOriented_Default()
...
+            {
+                int batchSize = Math.Min(WriteRowsPerGroup, values.Length - offset);
+
+                var indices = new int[batchSize];
+                var chunk = new float[batchSize];
+                for (int i = 0; i < batchSize; i++)
+                {
+                    indices[i] = offset + i;
+                    chunk[i] = values[offset + i];
+                }
```
adamreeve: We don't need to copy the values into a chunk, this can be simplified to:
```
var chunk = values.Slice(offset, batchSize);
```
(after values is changed to a Span as above)

docs/guides/WriteBenchmarks.md
```diff
@@ -0,0 +1,163 @@
...
+
+- **None** – Uncompressed. Baseline for size comparison.
+- **Snappy** – Fast compression, moderate ratio.
+- **Zstd** – Slower compression, higher ratio.
+
+---
+
+## Datasets
+
+All datasets are IEEE 754 32-bit single-precision (float) scientific data, decompressed from SPDP-compressed files.
```
adamreeve: This should include an inline link to the web page where these can be downloaded.

docs/guides/WriteBenchmarks.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Test Environment
+
+- **OS:** Ubuntu 24.04 LTS
+- **Execution:** VMware virtual machine
+- **RAM:** 32 GB
+- **CPU:** 2 vCPUs
+- **Storage:** Local file system
+- **Runtime:** `dotnet run --no-build --configuration=Release`
+- **Timing:** `/usr/bin/time -v`
```
adamreeve: Rather than using `/usr/bin/time` here, we should add timing inside the `ConvertData`, and only time from `using var writer = ...` to `write.Close()`.  Otherwise the timings include the time taken to read the input data, which is quite a lot of overhead.

docs/guides/WriteBenchmarks.md
```diff
@@ -0,0 +1,163 @@
...
+
+---
+
+## Results
+
+### num_control.sp.bin (76 MB raw · 19,938,093 floats)
+
+| Encoding | Compression | Parquet size (MB) | Wall time (s) | Peak RSS (MB) | Size vs Plain/None |
```
adamreeve: I don't think "Peak RSS" is as useful for the write benchmarks as it is for the read benchmarks. This will be dominated by the memory used to store the full input data. Maybe this can be removed.

It would be interesting to add the read time though, as different encodings and compressions will also affect read time and that trade off will be quite interesting to a lot of users.

docs/guides/WriteBenchmarks.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Analysis
+
+### Compression ratio
+
+ByteStreamSplit consistently produces the smallest Parquet files across all three datasets, reducing file size by **53–64%** compared to uncompressed Plain. This is expected behaviour: scientific float data has high entropy in the mantissa bytes but low entropy in the sign and exponent bytes. ByteStreamSplit separates these byte planes so the compressor sees long runs of similar bytes, dramatically improving the compression ratio.
+
+Plain + Zstd achieves a moderate improvement of **20–23%** over the uncompressed baseline, confirming that standard compression alone can compress float data but without byte-plane separation its gains are limited.
+
+Dictionary encoding is counterproductive for these datasets. Floats from scientific simulations and telescope observations are largely unique values; the dictionary overhead increases file sizes by **4–15%** compared to Plain with no dictionary. This is consistent with the expectation that dictionary encoding only helps when a column has a small number of repeating values.
```
adamreeve: Might be worth mentioning that dictionary encoding is enabled by default, so it's important to consider whether it should be disabled for floating point columns.

docs/guides/WriteBenchmarks.md
```diff
@@ -0,0 +1,163 @@
...
+| num_control | 19,938,093 | ~76 MB |
+| num_brain | 17,730,000 | ~67.6 MB |
+| obs_spitzer | 24,772,608 | ~94.5 MB |
+
+```
+Throughput (MB/s) = RawDataSize (MB) / Wall time (s)
+```
+
+---
+
```
adamreeve: This section looks like it can be removed, there aren't any throughput measurements mentioned elsewhere so I'm not sure what this is for.

adamreeve (commented):
Thanks @larrytamnjong, I've left some suggestions.

larrytamnjong:
@adamreeve I’ve resolved the suggestions you left, please take another look and let me know what you think.

docs/guides/WriteBenchmarks.md
```diff
@@ -0,0 +1,148 @@
...
+## Overview
+
+Three encoding strategies where tested, each paired with up to three compression codecs, across three real-world scientific float datasets. The datasets were first [decompressed](https://userweb.cs.txstate.edu/~burtscher/research/SPDPcompressor/) from SPDP format to raw binary (`.bin`) and then converted to Parquet.
+
+### Encodings tested
+
+- **Plain (no dictionary)** – Raw float values written as-is. No dictionary lookup overhead.
+- **Plain (dictionary enabled)** – Attempts to build a dictionary of unique float values per row group. Only beneficial if floats repeat frequently. Note that **dictionary encoding is enabled by default** in Parquet, so it is important to consider whether it should be disabled for floating-point columns where values are largely unique.
```
adamreeve:
```suggestion
- **Plain (dictionary enabled)** – Attempts to build a dictionary of unique float values per row group. Only beneficial if floats repeat frequently. Note that **dictionary encoding is enabled by default** in ParquetSharp, so it is important to consider whether it should be disabled for floating-point columns where values are largely unique.
```

adamreeve (commented):
These latest changes look good thanks Larry, I just have one small suggested change.

For the next step, it would be good to add measurements of how the write configuration affects the read time, when using consistent read options. Using the `logical-chunked` configuration would make sense as you found this was optimal in your previous tests. You might want to allow specifying the file to read when using the read benchmark commands to support this.

larrytamnjong:
@adamreeve I’ve added read measurements and moved the PR out of draft. Please let me know if everything looks okay.

larrytamnjong:
@adamreeve as discussed I have updated the read results to be an average of 5 runs.

adamreeve (approved):

adamreeve:
:+1: Thanks for all your work on this @larrytamnjong

PR #641, opened 2026-04-20
Fix documentation for `WriterPropertiesBuilder.MaxRowGroupLength`

### LINKED ISSUE #640: `WriterPropertiesBuilder.MaxRowGroupLength` refers to bytes when the bound method refers to rows
The documentation for `WriterPropertiesBuilder.MaxRowGroupLength` seems to incorrectly refer to the number of bytes in a RowGroup, when I think it is supposed to be rows.
https://github.com/G-Research/ParquetSharp/blob/7bf92fdca95a5e6068ff5a6c37fe18b64dfad68a/csharp/WriterPropertiesBuilder.cs#L346-L357

This eventually calls this native function:
https://github.com/G-Research/ParquetSharp/blob/7bf92fdca95a5e6068ff5a6c37fe18b64dfad68a/cpp/WriterPropertiesBuilder.cpp#L158-L161

which according to the C++ implementation documentation, is in terms of rows https://arrow.apache.org/docs/cpp/api/formats.html#_CPPv4N7parquet16WriterProperties7Builder20max_row_group_lengthE7int64_t

If this is actually the case, I can submit a PR to fix the documentation comments.

mobiusklein:
Fixes #640 

There is also some whitespace cleanup in the touched file. This might be a failure in my EditorConfig setup though.

### DIFF
--- a/csharp/WriterPropertiesBuilder.cs
+++ b/csharp/WriterPropertiesBuilder.cs
@@ -7,7 +7,7 @@ using ParquetSharp.Schema;
 namespace ParquetSharp
 {
     /// <summary>
-    /// Builder pattern for creating and configuring a <see cref="WriterProperties"/> object. 
+    /// Builder pattern for creating and configuring a <see cref="WriterProperties"/> object.
     /// </summary>
     public sealed class WriterPropertiesBuilder : IDisposable
     {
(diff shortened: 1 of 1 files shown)


marcin-krystianc (approved):
Looks good. Thanks!

PR #642, opened 2026-04-21
Add a limit on stackalloc size when reading Decimal columns

adamreeve:


csharp.test/TestDecimal.cs
```diff
@@ -434,6 +434,59 @@ public static void TestScaleMultiplier()
             Assert.AreEqual(1e+028M, DecimalConverter.GetScaleMultiplier(28, 29));
         }
 
+        // LogicalTypeFactory override that bypasses decimal precision calculation,
+        // and always assumes a decimal column with FLBA storage.
```
adamreeve, line 438: The precision calculation in `DecimalConverter.MaxPrecision` overflows and the cast to int is undefined, and the actual result changes between framework versions. So this wasn't necessary when I tested locally with .net10, but is needed for the test to run on .net8.

We could probably change `MaxPrecision` to avoid this overflow, but that seems out of scope for this fix.

marcin-krystianc (approved):

Issue #649, opened 2026-05-04
Use cached docker images for Linux CI builds


adamreeve:
See https://github.com/G-Research/ParquetSharp/pull/647/changes#r3180938421

We now build bison in the Docker container on every build. It would make sense to build a Docker image based on  the manylinux image that has the extra dependencies required to build ParquetSharp, and cache this to speed up CI builds.

PR #653, opened 2026-05-16
Use cached Docker images for Linux CI builds

MohamedM216:
issue: #649 

Currently, every Linux CI build compiles bison 3.8.2 from source inside a manylinux2014 container. This PR introduces a pre-built Docker image (ghcr.io/g-research/parquetsharp/linux-ci) based on manylinux2014 with bison and the full build toolchain baked in, published to GHCR via a new `build-docker.yml` workflow. The image is rebuilt automatically when the Dockerfile changes, weekly to pick up upstream updates, or on manual dispatch. Both amd64 and arm64 variants are built natively and combined into a multi-arch manifest. `ci.yml` is updated to pull this image instead of installing dependencies inline.

For more info see https://github.com/MohamedM216/ParquetSharp/actions/runs/25957299701
https://github.com/MohamedM216/ParquetSharp/actions/runs/25958080170
https://github.com/MohamedM216/ParquetSharp/actions/runs/25957406698
https://github.com/MohamedM216/ParquetSharp/pkgs/container/parquetsharp%2Flinux-ci

### DIFF
--- a/.github/docker/Dockerfile.linux-ci
+++ b/.github/docker/Dockerfile.linux-ci
@@ -1,0 +1,41 @@
+# syntax=docker/dockerfile:1
+#
+# CI image for building the ParquetSharp native library on Linux.
+#
+# Based on the manylinux2014 image so that the resulting shared library is
+# compatible with a wide range of Linux distributions (glibc >= 2.17).
+#
+# Bison and the rest of the build toolchain are baked in here so that CI
+# runners don't have to compile bison from source on every build.
+
+ARG ARCH=x86_64
+FROM quay.io/pypa/manylinux2014_${ARCH}
+
+ARG BISON_VERSION=3.8.2
+
+# Install SCL-based toolchain packages (devtoolset-10 = GCC 10, rh-git227, httpd24-curl)
+# plus flex and perl modules required by some vcpkg ports, then build bison from source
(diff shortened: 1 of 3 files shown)


MohamedM216:
Regarding the failing tests, we must run ` Build Linux CI Docker image` workflow first,  then trigger `CI` to run.

marcin-krystianc:
@MohamedM216, thanks a lot for the contribution and for tackling the issue. After some considerations, we've decided to go with #656 instead:

### This PR:
- Introduces a separate `build-docker.yml` workflow that must be run *before* `ci.yml`, creating an ordering dependency that's easy to get wrong
- Stores the image in GHCR as a long-lived artefact, which adds an external registry dependency that every CI run must pull from.
- The weekly rebuild schedule is nice, but it means the image can drift out of sync with the repo, and there's no guarantee the latest tag reflects the current `Dockerfile`.

### #656:
- Keeps everything self-contained in a single `ci.yml` workflow. No separate image publishing step, no ordering dependency, no external registry state to manage.
- Uses the native GitHub Actions cache (`type=gha`) for Docker layer caching. On a cache hit, the toolchain layers are restored in seconds without downloading anything from an external registry. On a cache miss, it rebuilds and stores the layers automatically.

MohamedM216:
No worries at all. The approach addressed in #656 appears to be significantly more performant.

PR #656, opened 2026-05-21
Enable caching when building CentOS container

### LINKED ISSUE #649: Use cached docker images for Linux CI builds

marcin-krystianc:
Fixes #649

Move the Linux CI build container setup from inline docker exec commands into a dedicated Dockerfile (`.github/ci-build/Dockerfile`) and use `docker/build-push-action` with GitHub Actions cache (`type=gha`) to cache the Docker image layers. This avoids reinstalling the toolchain (`devtoolset`, `bison`, `ninja`, etc.) on every CI run, speeding up the Linux build steps.


.github/ci-build/Dockerfile
```diff
@@ -0,0 +1,24 @@
+ARG MANYLINUX_ARCH
```
adamreeve: Can we set a default to avoid the Docker build warnings:
```suggestion
ARG MANYLINUX_ARCH=x86_64
```

adamreeve (commented):
Thanks Marcin, this looks good to me. Just one minor suggestion.

adamreeve (approved):

PR #694, opened 2026-08-10
chore: Don't force use of system binaries for vcpkg in devcontainer

adamreeve:
This is causing the devcontainer builds to fail because the version of CMake required by vcpkg is now more recent than what is available in Ubuntu 24.04.


adamreeve:
Closing as not needed, vcpkg was updated to work with older cmake

